PR #5, opened 2023-12-01
McStas loader and time binning workflow.

### LINKED ISSUE #6: Loaders for simulation data and measurement data.
There is a slight difference between simulation data and measurement data
In the measurement data, event data is provided per detector and simulation data has all event data in one place.
So for the measurement data, there is one more step to combine data sets from all detectors.

We also need to make small mcstas simulation data for testing.

YooSunYoung:
Fixes #6 
``docs/examples/workflow.ipynb`` has an example workflow that can load McStas data and bin into time dimension.

TODO:
- [x] Make small subset of simulation data for loader testing.
- [x] Update workflow example to use generated data or downloaded sample data.
~- [ ] Add another workflow example that has measurement data sample.~ Will be done separately.
- [x] Add unit tests for loaders and minimum reduction functions.
~- [ ] Add documentation of loaders.~


docs/examples/workflow.ipynb
```diff
@@ -0,0 +1,205 @@
...
+   "outputs": [],
+   "source": [
+    "from ess.nmx import build_workflow\n",
+    "from ess.nmx.loader import FileTypeMcStas\n",
+    "from ess.nmx.data import small_mcstas_sample\n",
+    "from ess.nmx.reduction import TimeBinned\n",
+    "\n",
+    "file_path = small_mcstas_sample()  # Replace it with your data file path\n",
+    "\n",
+    "nmx_workflow = build_workflow(file_path)\n",
```
nvaytet, line 56: In `esssans` and `essreflectometry`, we have kept the explicit use of `Pipeline` in the notebook, and we didn't hide it behind a `build_workflow` wrapper.

Should we keep it consistent and not use a wrapper?
Or do you think having to build the pipeline is too much code for users to write?

I personally think that it's good for users to have to build it, it gives them more understanding of what is going on under the hood.

But maybe it was a requirement from the NMX people?

src/ess/nmx/detector.py
```diff
@@ -0,0 +1,29 @@
+# SPDX-License-Identifier: BSD-3-Clause
+# Copyright (c) 2023 Scipp contributors (https://github.com/scipp)
+from dataclasses import dataclass
+from typing import NewType
+
+MaxNumberOfPixelsPerAxis = NewType("MaxNumberOfPixelsPerAxis", int)
+PixelStep = NewType("PixelStep", int)
+NumberOfDetectors = NewType("NumberOfDetectors", int)
+NumberOfAxis = NewType("NumberOfAxis", int)
```
nvaytet, line 9:
```suggestion
NumberOfAxes = NewType("NumberOfAxes", int)
```

src/ess/nmx/loader.py
```diff
@@ -0,0 +1,235 @@
...
+
+
+def get_events(
+    weights: Weights[FileType], t_list: TList[FileType], id_list: IDList[FileType]
+) -> Events[FileType]:
+    """get event list depending on dataformart and return dataset"""
+    return Events(sc.DataArray(data=weights, coords={'t': t_list, 'id': id_list}))
+
+
+providers = [
```
nvaytet, line 221: I think in other projects, the providers are now tuples instead of lists.

src/ess/nmx/loader.py
```diff
@@ -0,0 +1,235 @@
...
+def get_weights_mcstas(
+    loaded_data: LoadedData[FileTypeMcStas], max_prop: MaximumPropability
+) -> Weights[FileTypeMcStas]:
+    """Get weights of McStas data."""
+    weights: Weights = loaded_data['weights']
+
+    return Weights((max_prop / weights.max()) * weights)
+
+
+def get_t_list(loaded_data: LoadedData[FileTypeNMX]) -> TList[FileTypeNMX]:
```
nvaytet, line 174: Is this function missing a body? Or is it serving a different purpose?

src/ess/nmx/workflow.py
```diff
@@ -0,0 +1,47 @@
...
+from .reduction import TimeBinStep
+from .reduction import default_params as reduction_default_params
+from .reduction import providers as reduction_providers
+
+NMXWorkflow = NewType("NMXWorkflow", sl.Pipeline)
+NMXProviders = NewType("NMXProviders", list)
+NMXParams = NewType("NMXParams", dict)
+
+
+def collect_providers() -> NMXProviders:
```
nvaytet, line 20: Are the objects returned by the functions here used as part of a pipeline? If not, I think you don't need to wrap the return values into new types?

src/ess/nmx/loader.py
```diff
@@ -0,0 +1,235 @@
...
+
+InputFileName = NewType("InputFileName", str)
+
+FileType = TypeVar("FileType")
+FileTypeNMX = NewType("FileTypeNMX", str)
+FileTypeMcStas = NewType("FileTypeMcStas", str)
+
+# McStas Configurations
+MaximumPropability = NewType("MaximumPropability", int)
+DefaultMaximumPropability = MaximumPropability(100000)
```
nvaytet, line 20:
```suggestion
MaximumProbability = NewType("MaximumProbability", int)
DefaultMaximumProbability = MaximumProbability(100000)
```

src/ess/nmx/reduction.py
```diff
@@ -0,0 +1,132 @@
...
+
+providers = [
+    get_intervals_mcstas,
+    get_intervals,
+    get_ids,
+    get_grouped,
+    get_time_binned,
+]
+
+default_params = {TimeBinStep: DefaultTimeBinStep}
```
nvaytet, line 132: I would avoid having such a default parameter, and always ask the user to set it.
I think having default parameters for the detectors, such as their shape, is fine.

src/ess/nmx/workflow.py
```diff
@@ -0,0 +1,47 @@
...
+    return NMXProviders([*logging_providers, *loader_providers, *reduction_providers])
+
+
+def collect_default_parameters() -> NMXParams:
```
nvaytet, line 24: I think I would use the default parameters as in `essreflectometry`:
```Py
params={
    **default_parameters,
    QBins: sc.geomspace(dim='Q', start=0.008, stop=0.075, num=200, unit='1/angstrom'),
    SampleRotation[Sample]: sc.scalar(0.7989, unit='deg'),
    Filename[Sample]: "sample.nxs",
    SampleRotation[Reference]: sc.scalar(0.8389, unit='deg'),
    Filename[Reference]: "reference.nxs",
}
```
and then explicitly creating the pipeline in the notebook?

src/ess/nmx/reduction.py
```diff
@@ -0,0 +1,132 @@
...
+    interval_3: tuple
+
+
+class PixelIDs(sl.Scope[FileType, sc.Variable], sc.Variable):
+    """Pixel IDs for each detector"""
+
+    ...
+
+
+class Grouped(sl.Scope[FileType, sc.DataArray], sc.DataArray):
```
nvaytet, line 34: Could you make a name a little less generic? I didn't know what it was immediately. Something like `GroupedByDetectorId`? (if that is indeed what it is...)

src/ess/nmx/reduction.py
```diff
@@ -0,0 +1,132 @@
...
+
+    ...
+
+
+def calculate_distance(point_a: Vector3D, point_b: Vector3D) -> Distance:
+    """
+    Calculate the distance between two points.
+    """
+    diff = point_b - point_a
+    return Distance(sc.sqrt(sc.dot(diff, diff)))
```
nvaytet, line 51: Use `sc.norm`?

src/ess/nmx/reduction.py
```diff
@@ -0,0 +1,132 @@
...
+    ...
+
+
+class TimeBinned(sl.Scope[FileType, sc.DataArray], sc.DataArray):
+    """Time binned events"""
+
+    ...
+
+
+def calculate_distance(point_a: Vector3D, point_b: Vector3D) -> Distance:
```
nvaytet, line 46: This function appears to not be used anywhere?

src/ess/nmx/reduction.py
```diff
@@ -0,0 +1,132 @@
...
+    Calculate the distance between two points.
+    """
+    diff = point_b - point_a
+    return Distance(sc.sqrt(sc.dot(diff, diff)))
+
+
+RotationAngle = NewType("RotationAngle", sc.Variable)
+
+
+def rotation_matrix(axis: Vector3D, theta: RotationAngle) -> RotationMatirx:
```
nvaytet, line 57: Also unused?

src/ess/nmx/loader.py
YooSunYoung: Yes, the providers for ``FileTypeNMX`` could not be implemented yet... Should I just remove them...?

src/ess/nmx/reduction.py
YooSunYoung: Yes... they are from the original notebook. I didn't know how much would be included in the loader example. I removed them for now.

src/ess/nmx/workflow.py
YooSunYoung: Yes, I agree. I will remove it since providers are just tuples now.

src/ess/nmx/workflow.py
YooSunYoung: But it was a bit annoying to have ``default_parameters`` as a dictionary, because it's mutable.
I often accidentally changed the original dictionary so that is why I made this function instead...

src/ess/nmx/loader.py
SimonHeybrock, line 1: We should chat about all the providers and how things were split up here. In particular I don't think any of the McStas details should be exposed downstream, and none of the "downstream" types defined, e.g., in `reduction.py` should depend on the file type.

The domain types and providers in Sciline are intended for encapsulation and hiding implementation details of one part of the pipeline (such as loading data) for other parts (such as data reduction).

YooSunYoung:
Closing this in favor of #9

PR #7, opened 2023-12-11
Different version of loader.

YooSunYoung:
Another version of loader from #5.
This one doesn't have any Generic provider per file type.
Loading McStas file or nexus file should be handled automatically, but users can change parameters for loading McStas files.


YooSunYoung:
Closing this in favor of #9

PR #8, opened 2023-12-12
Copier update with dependencies.

YooSunYoung:
## Manually added configurations
- `pooch` as a base dependency.
- ~`ipywidget` as a dependency of docs.~  **Decided not to include this for now.**


SimonHeybrock:
>     * `ipywidget` as a dependency of docs.

Why is this needed? For the logger (which we decided to drop for now)?

YooSunYoung:
> > ```
> > * `ipywidget` as a dependency of docs.
> > ```
> 
> Why is this needed? For the logger (which we decided to drop for now)?

That is right.. I'll remove it...! Thanks!

SimonHeybrock (approved):

PR #9, opened 2023-12-12
McStas data loader and minimal data reduction as an example.

YooSunYoung:
Related to #6.


src/ess/nmx/loader.py
SimonHeybrock, line 1: Should we rename this file to `mcstas_loader.py` or something like that?

src/ess/nmx/loader.py
```diff
@@ -0,0 +1,92 @@
+# SPDX-License-Identifier: BSD-3-Clause
+# Copyright (c) 2023 Scipp contributors (https://github.com/scipp)
+from dataclasses import dataclass
+from typing import Iterable, NewType, Optional
+
+import scipp as sc
+import scippnexus as snx
+
+PixelIDs = NewType("PixelIDs", sc.Variable)
+InputFileName = NewType("InputFileName", str)
```
SimonHeybrock, line 10:
```suggestion
InputFilename = NewType("InputFilename", str)
```

src/ess/nmx/loader.py
```diff
@@ -0,0 +1,92 @@
...
+            return key
+    raise ValueError("Can not find event list name.")
+
+
+def _copy_partial_var(
+    var: sc.Variable, dim: str, idx: int, unit: str, dtype: Optional[str] = None
+) -> sc.Variable:
+    """Retrieve property from variable."""
+    original_var = var[dim, idx]
+    var = original_var.copy().astype(dtype) if dtype else original_var.copy()
```
SimonHeybrock, line 30:
```suggestion
    original_var = var[dim, idx].copy()
    var = original_var.astype(dtype) if dtype else original_var
```
?

src/ess/nmx/loader.py
```diff
@@ -0,0 +1,92 @@
...
+    McStas simulation file should have an entry path called ``entry1``.
+
+    Pixel id coordinate are wrapped together with the data
+    since they are different for simulation and real data.
+    """
+    with snx.File(file_name) as file:
+        if "entry1" in file:  # McStas file
+            return _load_mcstas_nmx_file(file, max_prop or DefaultMaximumProbability)
+        elif 'entry' in file:  # Nexus file
+            raise NotImplementedError("Measurement data loader is not implemented yet.")
```
SimonHeybrock, line 90: This is not going to work. "ESS" NeXus files may also have an entry named "entry1" (naming is arbitrary). I think we should not attempt to auto-detect the file. Simply remove everything that is not for McStas. Name the function `load_mcstas` or something like that?

Not sure it matters here, but you could also look for a unique top-level entry (this is the most common case), instead of a fixed name.

src/ess/nmx/loader.py
```diff
@@ -0,0 +1,92 @@
...
+    from functools import partial
+
+    bank_name = _retrieve_event_list_name(file["entry1/data"].keys())
+    var: sc.Variable
+    var = file["entry1/data/" + bank_name]["events"][()].rename_dims({'dim_0': 'event'})
+
+    copier = partial(_copy_partial_var, var, dim='dim_1')
+    weights = copier(idx=0, unit='counts')
+    id_list = copier(idx=4, unit='dimensionless', dtype="int64")
+    t_list = copier(idx=5, unit='s')
```
SimonHeybrock, line 64: I don't feel the code got more readable by having `partial` here. Can you just call the function directly? The `dim_1` might just as well be hard-coded, it doesn't seem like there are any other cases?

src/ess/nmx/loader.py
```diff
@@ -0,0 +1,92 @@
...
+    """
+    from functools import partial
+
+    bank_name = _retrieve_event_list_name(file["entry1/data"].keys())
+    var: sc.Variable
+    var = file["entry1/data/" + bank_name]["events"][()].rename_dims({'dim_0': 'event'})
+
+    copier = partial(_copy_partial_var, var, dim='dim_1')
+    weights = copier(idx=0, unit='counts')
+    id_list = copier(idx=4, unit='dimensionless', dtype="int64")
```
SimonHeybrock, line 63: IDs should never have a unit, use `unit=None`, otherwise a bunch of operations may not work as expected.

src/ess/nmx/loader.py
```diff
@@ -0,0 +1,92 @@
...
+    t_list = copier(idx=5, unit='s')
+
+    weights = (max_prop / weights.max()) * weights
+
+    loaded = sc.DataArray(data=weights, coords={'t': t_list, 'id': id_list})
+    pixel_ids = _get_mcstas_pixel_ids()
+    return NMXData(Events(loaded), pixel_ids)
```
SimonHeybrock, line 70: Why the helper class? Shouldn't we group directly here?

src/ess/nmx/loader.py
```diff
@@ -0,0 +1,92 @@
...
+Events = NewType("Events", sc.DataArray)
+
+# McStas Configurations
+MaximumProbability = NewType("MaximumProbability", int)
+DefaultMaximumProbability = MaximumProbability(100_000)
+
+
+def _retrieve_event_list_name(keys: Iterable[str]) -> str:
+    for key in keys:
+        if key.startswith("bank01_events_dat_list"):
```
SimonHeybrock, line 20: Should we include the list of required fields (which I think was also encoded in the group name)?

src/ess/nmx/reduction.py
```diff
@@ -0,0 +1,21 @@
...
+
+def get_grouped_by_pixel_id(
+    loaded: NMXData, num_panels: NumberOfDetectors
+) -> GroupedByPixelID:
+    """group events by pixel ID"""
+    grouped = loaded.events.group(loaded.all_pixel_ids)
+
+    return GroupedByPixelID(
+        grouped.fold(dim='id', sizes={'panel': num_panels, 'id': -1})
+    )
```
SimonHeybrock, line 21: This grouping is McStas specific, move into the load function (ScippNeXus returns grouped events already).

src/ess/nmx/detector.py
```diff
@@ -0,0 +1,16 @@
...
+NumberOfDetectors = NewType("NumberOfDetectors", int)
+NumberOfAxes = NewType("NumberOfAxes", int)
+
+
+default_params = {
+    MaxNumberOfPixelsPerAxis: MaxNumberOfPixelsPerAxis(1280),
+    NumberOfAxes: NumberOfAxes(2),
+    PixelStep: PixelStep(1),
+    NumberOfDetectors: NumberOfDetectors(3),
+}
```
SimonHeybrock, line 16: Is all this required? Can it not be detector from the input data? It seems the McStas loader is anyway using hard-coded 3 detectors.

YooSunYoung:
@Justin-Bergmann 

We have a couple of questions for you about this McStas loader...!

## Will you ever need non-grouped data?
If not, we will just always group them into pixel IDs when you load a file.

## Are p_x_y_n_id_t these variables always in the McStas simulation? 
We would like to know what's the minimum requirements of NMX McStas simulation data event list name.
For example, if the event data path in the h5 file ``entry1/data/bank01_events_dat_list_p_x_y_n_id_t`` must contain at least _p_ _id_ _t_ .

And we're going to hard-code the translation of each field, for example `p` is always `weights`.
So please let us know if the names might change in the McStas files...!

src/ess/nmx/loader.py
YooSunYoung: > This is not going to work. "ESS" NeXus files may also have an entry named "entry1" (naming is arbitrary). I think we should not attempt to auto-detect the file. Simply remove everything that is not for McStas. Name the function load_mcstas or something like that?

Yes, I removed that part. So users should just use different provider for loading depending on the file...? 

> Not sure it matters here, but you could also look for a unique top-level entry (this is the most common case), instead of a fixed name.

I don't what you mean by unique top-level entry...

src/ess/nmx/detector.py
YooSunYoung: Yes... they should be included in the file as well... I was procrastinating parsing the detector information...

Justin-Bergmann: in principle, we can have more or less  detector panels

Justin-Bergmann:
> @Justin-Bergmann
> 
> We have a couple of questions for you about this McStas loader...!
> 
> ## Will you ever need non-grouped data?
> If not, we will just always group them into pixel IDs when you load a file.
> 
> ## Are p_x_y_n_id_t these variables always in the McStas simulation?
> We would like to know what's the minimum requirements of NMX McStas simulation data event list name. For example, if the event data path in the h5 file `entry1/data/bank01_events_dat_list_p_x_y_n_id_t` must contain at least _p_ _id_ _t_ .
> 
> And we're going to hard-code the translation of each field, for example `p` is always `weights`. So please let us know if the names might change in the McStas files...!

Spontaneously, I can not imagine any situation where we would use non-grouped data and we still have all information available. So please go ahead.

Yes, each simulation must contain p, id and t and as far as I know  all p_x_y_n_id_t are always in the simulation

The names will never change as long its not changed in the code itself

docs/examples/workflow.ipynb
```diff
@@ -8,9 +8,9 @@
     "\n",
     "## Collect Parameters and Providers\n",
     "### Simulation(McStas) Data\n",
-    "There is an extra parameter that can be configured for **McStas** simulation data workflow. <br>\n",
-    "It is because ``weights`` are given as probability, not integer number of events in a McStas file. <br>\n",
-    "Therefore users can set a scale factor called ``MaximumProabability`` to derive more realistic number of events. <br>"
+    "There is a dedicated loader, ``load_mcstas_nmx_file`` for ``McStas`` simulation data workflow. <br>\n",
+    "``MaximumProabability`` can be manually provided to the loader derive more realistic number of events. <br>\n",
```
SimonHeybrock, line 12: Spelling.

Also, since this is in the "nmx" package, can we drop the "nmx" from the function name? I'd suggest `load_mcstas_nexus`, for example (since we cannot load other McStas files, only NeXus).

src/ess/nmx/loader.py
SimonHeybrock: > I don't know what you mean by unique top-level entry...

Ever NeXus file has one or more "entries" at the file root. Most files have just one, but the name is arbitrary.

src/ess/nmx/mcstas_loader.py
```diff
@@ -0,0 +1,96 @@
...
+
+    raise ValueError("Can not find event list name.")
+
+
+def _copy_partial_var(
+    var: sc.Variable, idx: int, unit: Optional[str] = None, dtype: Optional[str] = None
+) -> sc.Variable:
+    """Retrieve a property from a variable."""
+    original_var = var['dim_1', idx].copy()
+    var = original_var.astype(dtype) if dtype else original_var
```
SimonHeybrock, line 43:
```suggestion
    var = original_var.astype(dtype, copy=False) if dtype else original_var
```

src/ess/nmx/__init__.py
```diff
@@ -10,3 +10,27 @@
...
+
+__all__ = [
+    "small_mcstas_sample",
+    "NumberOfDetectors",
+    "load_mcstas_nexus",
+    "NMXData",
+    "InputFilepath",
+    "PixelIDs",
+    "MaximumProbability",
+    "DefaultMaximumProbability",
```
SimonHeybrock, line 34: I think we should keep McStas-load specific params in the McStas module, and not import on the root level?

src/ess/nmx/mcstas_loader.py
```diff
@@ -0,0 +1,96 @@
...
+        weights = _copy_partial_var(var, idx=0, unit='counts')
+        id_list = _copy_partial_var(var, idx=4, dtype='int64')
+        t_list = _copy_partial_var(var, idx=5, unit='s')
+
+        weights = (prop / weights.max()) * weights
+
+        loaded = sc.DataArray(data=weights, coords={'t': t_list, 'id': id_list})
+        grouped = loaded.group(_get_mcstas_pixel_ids())
+
+        return NMXData(grouped.fold(dim='id', sizes={'panel': panels, 'id': -1}))
```
SimonHeybrock, line 96: This looks broken since pixel-IDs are always for 3 panels? I would suggest to simply remove the panels option. I think we will simply get empty event lists in the case?

src/ess/nmx/mcstas_loader.py
```diff
@@ -0,0 +1,96 @@
...
+
+def _retrieve_event_list_name(keys: Iterable[str]) -> str:
+    prefix = "bank01_events_dat_list"
+    mandatory_keys = ('_p', '_x', '_y', '_n', '_id', '_t')
+
+    for key in keys:
+        if key.startswith(prefix) and _find_all(
+            key.removeprefix(prefix), *mandatory_keys
+        ):
+            return key
```
SimonHeybrock, line 33: You seem to be checking for keys in arbitrary order here? Below you are using hard-coded column indices. Before making such a generalization, consider carefully:

- Is it actually required?
- What can break if I generalize?
- Is it worth the effort (including writing unit tests for the generalization!)?

In this case I think the generalization should be removed, just check for the known fixed order.

SimonHeybrock (changes requested):

src/ess/nmx/detector.py
```diff
@@ -0,0 +1,16 @@
...
+PixelStep = NewType("PixelStep", int)
+NumberOfDetectors = NewType("NumberOfDetectors", int)
+NumberOfAxes = NewType("NumberOfAxes", int)
+
+
+default_params = {
+    MaxNumberOfPixelsPerAxis: MaxNumberOfPixelsPerAxis(1280),
+    NumberOfAxes: NumberOfAxes(2),
+    PixelStep: PixelStep(1),
+    NumberOfDetectors: NumberOfDetectors(3),
```
SimonHeybrock, line 15: Partially unused now?

YooSunYoung: I think none of them are used actually... I'll just remove this module.

src/ess/nmx/mcstas_loader.py
```diff
@@ -0,0 +1,82 @@
...
+
+    raise ValueError("Can not find event list name.")
+
+
+def _copy_partial_var(
+    var: sc.Variable, idx: int, unit: Optional[str] = None, dtype: Optional[str] = None
+) -> sc.Variable:
+    """Retrieve a property from a variable."""
+    original_var = var['dim_1', idx].copy()
+    var = original_var.astype(dtype, copy=False) if dtype else original_var
```
SimonHeybrock, line 35: One final question. Can we use
```suggestion
    var = var['dim_1', idx].astype(dtype or var.dtype , copy=True)
```
to avoid up to one copy?

src/ess/nmx/mcstas_loader.py
```diff
@@ -0,0 +1,81 @@
...
+def _get_mcstas_pixel_ids() -> PixelIDs:
+    """pixel IDs for each detector"""
+    intervals = [(1, 1638401), (2000001, 3638401), (4000001, 5638401)]
+    ids = [sc.arange('id', start, stop, unit=None) for start, stop in intervals]
+    return PixelIDs(sc.concat(ids, 'id'))
+
+
+def load_mcstas_nexus(
+    file_path: InputFilepath,
+    max_prop: Optional[MaximumProbability] = None,
```
SimonHeybrock, line 49:
```suggestion
    max_prob: Optional[MaximumProbability] = None,
```
or do not abbreviate?

SimonHeybrock (approved):

PR #14, opened 2023-12-21
Update dependencies.

YooSunYoung:
# Added dependencies.
- pythreejs: For instrument views.
- defusedxml: For parsing xml(instrument geometry).
- scippneutron: For instrument views (only in docs)

# TODO:
- [x] Build nightly dependency again.
**Nightly dependency could not be compiled since I was using mac os and my workstation is offline...**


pyproject.toml
```diff
@@ -37,6 +37,7 @@ dependencies = [
   "scipp>=23.8.0",
   "scippnexus>=23.9.0",
   "pooch",
+  "defusedxml",
```
SimonHeybrock, line 40: Can you explain the choice of a third-party library over something builtin in Python? Do you consider the data we get untrusted?

YooSunYoung: This one was recommended by the bandit check because of a security vulnerability of builtin package for xml...
Builtin package xml was not allowed when I tried to commit the code.

SimonHeybrock: 👍

SimonHeybrock (approved):

Issue #16, opened 2023-12-22
Save results into nexus format.


YooSunYoung:
The result of first data reduction step should be saved into nexus format to be fed to the next process.

It can be implemented once #15 is finished.

# Required Fields
- [x] Sample name
- [x] Distance from source to sample
- [x] Other source information
- [x] Detector origen(?)
- [x] Fast axes (in beamline frame(?))
- [x] Slow axes (in beamline frame(?))
- [x] Proton charge (?)
- [x] Detector counts (grouped weights/counts, gzip compressed)
- [x] Time bins (in [ms])
- [x] Pixel IDs
- [x] Crystal Rotation

# Questions
- Is there anything can be already done in the first reduction step instead...?
- Many fields used to be hard-coded and not straightforward where to retrieve them from the McStas file.
  Should users be able to update them manually some how...?

YooSunYoung:
Here are more detailed explanations of each field.

- Other source information
   Mostly hardcoded/manually input except for the number of detectors.
   All other information related to each detector should be in a combined list, sorted by the detector ID.)
- Detector origen(?)
  3d vector position of the first pixel from the sample
- Fast axes
   The axis along which pixel IDs grows by the number of pixels in one row (1280)
- Slow axes
   The axis along which pixel IDs grows by 1
- Proton charge
   Number of protons hitting the target within the measurement time frames
   It is proportional to the normalization factor. Normalization is done on the integrated events.
   For simulation data, it doesn't have to be precise, and can be arbitrarily derived by the number of events.
   (Since McStas simulation starts from neutron, not proton.)
- Time bins (in [ms])
   Edges of time binned data.

PR #15, opened 2023-12-21
Mcstas instrument parser and data loader.

YooSunYoung:
Fixes #10 
Data loader now load geometry coordinates as well by parsing instrument geometry xml in the McStas file.\

TODO:
- [x] Write algebra tests.
- [ ] Write position checking tests.

PR #18, opened 2024-01-09
Mcstas instrument geometry in the loader.

### LINKED ISSUE #10: Geometric information parsing (McStas loader mainly)
NMX data loader should be able to retrieve...

- detector rotation (x, y, z)
- detector position
- relative (x, y, z) list (?)
- source position
- sample position
- fast vector list
- slow vector list

In the McStas simulation file, there is a h5 group ``instrument/instrument_xml/`` that contains geometry information of the instruments.
McStas loader should be able to parse the xml information to retrieve the information above.

TODO:
 - [ ] Compare instrument view with mantid

YooSunYoung:
Fixes #10 

Continued from #15 

Now all fields that could be derived from the geometry xml description are included as coordinates of the ``NMXData`` (the output type of the workflow.)

image

These coordinates cover most of the output fields that need to be saved as an intermediate result for #16

### DIFF
--- a/src/ess/nmx/mcstas_xml.py
+++ b/src/ess/nmx/mcstas_xml.py
@@ -8,7 +8,6 @@ from typing import Iterable, Optional, Protocol, Tuple, TypeVar, Union
 
 import numpy as np
 import scipp as sc
-from typing_extensions import Self
 
 T = TypeVar('T')
 
@@ -31,13 +30,13 @@ class _XML(Protocol):
     tag: str
     attrib: dict[str, str]
 
-    def find(self, name: str) -> Optional[Self]:
+    def find(self, name: str) -> Optional['_XML']:
         ...
 
-    def __iter__(self) -> Self:
+    def __iter__(self) -> '_XML':
         ...
 
-    def __next__(self) -> Self:
+    def __next__(self) -> '_XML':
         ...
 
 
@@ -94,7 +93,7 @@ class SimulationSettings:
     handedness: str  # 'val' of <handedness>
 
     @classmethod
-    def from_xml(cls, tree: _XML) -> Self:
+    def from_xml(cls, tree: _XML) -> 'SimulationSettings':
         """Create simulation settings from xml."""
         defaults = select_by_tag(tree, 'defaults')
         length_desc = select_by_tag(defaults, 'length')
@@ -157,7 +156,7 @@ class DetectorDesc:
     @classmethod
     def from_xml(
         cls, component: _XML, type_desc: _XML, simulation_settings: SimulationSettings
-    ) -> Self:
+    ) -> 'DetectorDesc':
         """Create detector description from xml component and type."""
 
         def _rotate_axis(matrix: sc.Variable, axis: sc.Variable) -> sc.Variable:
@@ -246,7 +245,9 @@ class SampleDesc:
     rotation_matrix: sc.Variable
 
     @classmethod
-    def from_xml(cls, tree: _XML, simulation_settings: SimulationSettings) -> Self:
+    def from_xml(
+        cls, tree: _XML, simulation_settings: SimulationSettings
+    ) -> 'SampleDesc':
         """Create sample description from xml component."""
         source_xml = select_by_type_prefix(tree, 'sampleMantid-type')
         location = select_by_tag(source_xml, 'location')
@@ -286,7 +287,9 @@ class SourceDesc:
     position: sc.Variable
 
     @classmethod
-    def from_xml(cls, tree: _XML, simulation_settings: SimulationSettings) -> Self:
+    def from_xml(
+        cls, tree: _XML, simulation_settings: SimulationSettings
+    ) -> 'SourceDesc':
         """Create source description from xml component."""
         source_xml = select_by_type_prefix(tree, 'sourceMantid-type')
         location = select_by_tag(source_xml, 'location')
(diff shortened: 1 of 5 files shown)


YooSunYoung:
The instrument view is inserted as a code-snippet not as a cell.
It was because the output of the instrument view causes the error/warnings below
and the notebook is not linked to the ``index.html`` at the end, since it couldn't find the title...

```
ERROR: Line 404 exceeds the line-length-limit.
WARNING: Each notebook should have at least one section title
```

src/ess/nmx/mcstas_xml.py
```diff
@@ -0,0 +1,396 @@
...
+            position=_position_from_location(location, simulation_settings.length_unit),
+            rotation_matrix=_rotation_matrix_from_location(
+                location, simulation_settings.angle_unit
+            ),
+        )
+
+    def position_from_sample(self, other: sc.Variable) -> sc.Variable:
+        """Position of ``other`` relative to the sample.
+
+        All positions and distance are stored respect to the sample position.
```
SimonHeybrock, line 267:
```suggestion
        All positions and distance are stored relative to the sample position.
```

src/ess/nmx/mcstas_xml.py
```diff
@@ -0,0 +1,396 @@
...
+        )
+
+
+def _construct_pixel_ids(detector_descs: Tuple[DetectorDesc, ...]) -> sc.Variable:
+    """Pixel IDs for all detectors."""
+    intervals = [
+        (desc.id_start, desc.id_start + desc.total_pixels) for desc in detector_descs
+    ]
+    ids = [sc.arange('id', start, stop, unit=None) for start, stop in intervals]
+    return sc.concat(ids, 'id')
```
SimonHeybrock, line 310: Are we using a flat list of IDs? Why not a "panel" dim, or a DataGroup?

src/ess/nmx/mcstas_xml.py
```diff
@@ -0,0 +1,396 @@
...
+
+def _pixel_positions(
+    detector: DetectorDesc, position_offset: sc.Variable
+) -> sc.Variable:
+    """Position of pixels of the ``detector``.
+
+    Position of each pixel is relative to the position_offset.
+    """
+    pixel_idx = sc.arange('id', detector.total_pixels)
+    n_row = sc.scalar(detector.num_fast_pixels_per_row)
```
SimonHeybrock, line 321: Isn't this `n_col`?

src/ess/nmx/mcstas_xml.py
```diff
@@ -0,0 +1,396 @@
...
+        origins = [
+            self.sample.position_from_sample(det.position) for det in self.detectors
+        ]
+        detector_dim = 'panel'
+
+        return {
+            'pixel_ids': _construct_pixel_ids(self.detectors),
+            'fast_axis': sc.concat(fast_axes, detector_dim),
+            'slow_axis': sc.concat(slow_axes, detector_dim),
+            'origin_position': sc.concat(origins, detector_dim),
```
SimonHeybrock, line 380: This is the origin of each panel?

src/ess/nmx/mcstas_xml.py
```diff
@@ -0,0 +1,396 @@
...
+        """Extract coordinates from the McStas instrument description."""
+        slow_axes = [det.slow_axis for det in self.detectors]
+        fast_axes = [det.fast_axis for det in self.detectors]
+        origins = [
+            self.sample.position_from_sample(det.position) for det in self.detectors
+        ]
+        detector_dim = 'panel'
+
+        return {
+            'pixel_ids': _construct_pixel_ids(self.detectors),
```
SimonHeybrock, line 377:
```suggestion
            'pixel_id': _construct_pixel_ids(self.detectors),
```
we generally go with singular names.

src/ess/nmx/mcstas_xml.py
YooSunYoung: Because we have to group by the ids fist and I thought group by doesn't work with 2 dimensional array....?

src/ess/nmx/mcstas_xml.py
YooSunYoung: Yes it is. The position of the first pixel (lowest pixel ID).
I couldn't come up with a better name...

src/ess/nmx/mcstas_xml.py
YooSunYoung: Yeah actually column is more correct.

src/ess/nmx/mcstas_xml.py
```diff
@@ -0,0 +1,396 @@
...
+    slow_axis: sc.Variable
+
+    @classmethod
+    def from_xml(
+        cls, component: _XML, type_desc: _XML, simulation_settings: SimulationSettings
+    ) -> 'DetectorDesc':
+        """Create detector description from xml component and type."""
+
+        def _rotate_axis(matrix: sc.Variable, axis: sc.Variable) -> sc.Variable:
+            return sc.vector(np.round((matrix * axis).values, 2))
```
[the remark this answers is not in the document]
YooSunYoung: Justin said that's the precision that is needed so I kept it here.
Hmm... but I will remove rounding here so that it is handled in the next step if necessary.

src/ess/nmx/mcstas_xml.py
SimonHeybrock: Right!

src/ess/nmx/mcstas_xml.py
```diff
@@ -0,0 +1,410 @@
...
+        cls,
+        *,
+        component: _XML,
+        type_desc: _XML,
+        simulation_settings: SimulationSettings,
+    ) -> 'DetectorDesc':
+        """Create detector description from xml component and type."""
+
+        def _rotate_axis(matrix: sc.Variable, axis: sc.Variable) -> sc.Variable:
+            return matrix * axis
```
SimonHeybrock, line 169: This helper seems pointless now, just inline the multiplication?

SimonHeybrock (approved):

PR #19, opened 2024-01-09
Logos and icons.

YooSunYoung:
Static logo and icon images for documentation pages.

- [x] favicon.cio
- [x] logo.svg (text converted to path)
- [x] logo-dark.svg (text converted to path)
- [x] logo-text.svg (text as text with font indicated)


docs/_static/logo-text.svg
SimonHeybrock, line 1: Looks like wrong font on this one?

docs/_static/logo.svg
SimonHeybrock, line 1: Generally we keep the non-path version for enabling future modifications, typically in `resources/`. Why did you add it here? I think the docs use only two files, right?

SimonHeybrock (changes requested):

docs/_static/logo.svg
YooSunYoung: Yes, sorry...! I will relocate this file.

docs/_static/logo-text.svg
YooSunYoung: I just copied scipp logo and made them into path though...?

SimonHeybrock (approved):

PR #21, opened 2024-01-11
Time binning reduction and export method.

### LINKED ISSUE #3: Data reduction workflow skeleton for NMX.
NMX already has a working script for data reduction written by the IDS owner(Justin).
We can build workflow with sciline from the script.

### LINKED ISSUE #16: Save results into nexus format.

YooSunYoung:
Fixes #16
Fixes #3 

# Main Changes
## Exporting reduced data as nexus file.
The exporting method was implemented together with the time binning so that it is more clear what to save into the result.

## Smaller steps as parameters.
``event weights converter`` and ``proton_charge_from_event_data`` are set as parameters.
This solution minimizes the fraction of the McStas specific part in the workflow graph,
yet allow users to access to the smaller steps of the McStas loader.

The reason of having them as parameters not as providers is,
- They are not part of general reduction, which are only for McStas cases.
And we wanted to keep the McStas specific parts as compact as possible, i.e. not exposing all the magic numbers as parameters. Also because it'll be almost never changed.

- They are better done while the file is open and read in the loader.
This prevented splitting the loader function into smaller steps.

## TODO
- [x] Update instrument view scale unit to ``counts``.
- [x] Write tests of the export method.
- [x] Write tests for the crystal rotation angle. 
- [x] Write tests for all necessary fields
- [x] ~Use chexus (?)~ *I'll take this out as a separate issue*


YooSunYoung:
I've got another idea of how to deal with smaller step so I'll put it to draft for now...!

jl-wynen:
> I've got another idea of how to deal with smaller step so I'll put it to draft for now...!

I'm not familiar with this code or how it used to work. Can you point me to your solution for this in particular?

src/ess/nmx/mcstas_loader.py
```diff
@@ -27,54 +39,172 @@ def _retrieve_event_list_name(keys: Iterable[str]) -> str:
...
+    # Arbitrary number to scale the proton charge
+    _proton_charge_scale_factor = sc.scalar(1 / 10_000, unit=None)
+
+    return ProtonCharge(_proton_charge_scale_factor * event_da.bins.size().sum().data)
+
+
 def load_mcstas_nexus(
     file_path: InputFilepath,
+    event_weights_converter: McStasEventWeightsConverter,
+    proton_charge_converter: McStasProtonChargeConverter,
```
YooSunYoung, line 153: > > I've got another idea of how to deal with smaller step so I'll put it to draft for now...!
> 
> I'm not familiar with this code or how it used to work. Can you point me to your solution for this in particular?

It's this part I added.
It's explained in more detail here on the top:  https://github.com/scipp/essnmx/pull/21#issue-2076197190

src/ess/nmx/mcstas_loader.py
```diff
@@ -1,18 +1,30 @@
...
-NMXData = NewType("NMXData", sc.DataGroup)
 
 # McStas Configurations
 MaximumProbability = NewType("MaximumProbability", int)
 DefaultMaximumProbability = MaximumProbability(100_000)
 
+ConvertedEventWeights = NewType("ConvertedEventWeights", sc.Variable)
+McStasEventWeightsConverter = Callable[..., ConvertedEventWeights]
+# It should be ``Callable[[MaximumProbability, sc.Variable], ConvertedEventWeights]``
+# but sciline pipeline breaks if there is a list in the type arguments.
```
SimonHeybrock, line 20: Use `NewType`?

src/ess/nmx/mcstas_loader.py
```diff
@@ -27,54 +39,172 @@ def _retrieve_event_list_name(keys: Iterable[str]) -> str:
...
+
+    return sc.vector(
+        value=[file[f"entry1/simulation/Param/XtalPhi{key}"][...] for key in "XYZ"],
+        unit=unit,
+    )
+
+
+def event_weights_from_probability(
+    max_probability: MaximumProbability, probabilities: sc.Variable
+) -> ConvertedEventWeights:
```
SimonHeybrock, line 72: How about renaming this to `EventWeights`? The "converted" is a bit meaningless, and it may be easier to find a good name for the "unconverted" ones (such as `McStasWeights`, `RawProbability`, ...).

src/ess/nmx/mcstas_loader.py
```diff
@@ -27,54 +39,172 @@ def _retrieve_event_list_name(keys: Iterable[str]) -> str:
...
+    and the number of neutrons is not included in the result.
+
+    Parameters
+    ----------
+    event_da:
+        The event data binned in detector panel and pixel id dimensions.
+
+    """
+    # Arbitrary number to scale the proton charge
+    _proton_charge_scale_factor = sc.scalar(1 / 10_000, unit=None)
```
SimonHeybrock, line 145: Shouldn't there be more protons than neutrons?

src/ess/nmx/mcstas_loader.py
```diff
@@ -27,54 +39,172 @@ def _retrieve_event_list_name(keys: Iterable[str]) -> str:
...
+        The number of (detector) panels used in the experiment.
+
+    """
+
+    events = sc.DataArray(data=weights, coords={'t': t_list, 'id': id_list})
+    grouped: sc.DataArray = events.group(pixel_ids)
+    return grouped.fold(dim='id', sizes={'panel': num_panels, 'id': -1})
+
+
+def proton_charge_from_event_data(event_da: sc.DataArray) -> ProtonCharge:
```
SimonHeybrock, line 127: Not sure which `event_da` is passed here, but shouldn't it be a monitor instead of detector data? Otherwise we include the scattering effects, which may be problematic?

src/ess/nmx/mcstas_loader.py
SimonHeybrock: This pretty much makes this function unusable to load a file "by hand", without a workflow. Is this the simplest option, and does it solve an *actual* problem?

YooSunYoung: Why not...?
All the converters are included in the package so you can just inject the functions by hand.

Or we can make them `Optional` and set default values. Should we...?

src/ess/nmx/mcstas_loader.py
YooSunYoung: Yes, I asked about this before.
Justin said there should be easier/more precise ways to integrate number of protons during the measurements in the real experiments...
So, for simulation data, this `proton_charge` just need to be proportional to the number of events.
And the Monitor is not included in the simulation data... so it was the only way...?

src/ess/nmx/mcstas_loader.py
YooSunYoung: I just tried using ``NewType`` instead, and then documentation generation fails.
It doesn't throw any errors but it doesn't have any details either... 

image

YooSunYoung: Wait maybe it's not related...

YooSunYoung: Right...! It was not related, sorry...! I fixed it!

src/ess/nmx/mcstas_loader.py
SimonHeybrock: I meant UX wise.

Yes, there should be defaults.

src/ess/nmx/mcstas_loader.py
YooSunYoung: Hmm... I don't remember... 
@Justin-Bergmann should this scale factor be `> 1` ...?

src/ess/nmx/mcstas_loader.py
SimonHeybrock: Ok!

src/ess/nmx/mcstas_loader.py
YooSunYoung: It doesn't fix real problems, but it does fix the readability/accessibility problems,
to avoid duplicating/verbose documentation and hidden smaller steps in the loader.

In this way, each steps are easily exposed even though they are smaller steps in the loader to users
and we don't have to duplicate all the documentation what each step does.
You could avoid it by not separating those steps into smaller functions,
but then it becomes very verbose and less readable.

SimonHeybrock (approved):

PR #23, opened 2024-01-18
Fix instrument view in docs

nvaytet:
Do not show all pixels to reduce docs output size and fix error with line being too long.

Note that the data is just noise on the panels. Is that intentional?


YooSunYoung:
> Note that the data is just noise on the panels. Is that intentional?

You mean the data we're using for visualization? Yes, we're not using any sample data that represents certain experiment...

PR #30, opened 2024-02-09
Rename mcstas loader

YooSunYoung:
Updating McStas version from 2 to 3 will break the loader.
Therefore IDS(Justin) agreed with specifying the version of the McStas that current loader is expecting
in the documentation and the function name.
Justin said it will be much later that NMX is going to use McStas 3,
so we are not expecting any urgent request of McStas version 3 loader right now.

If general McStas loader is implemented as mentioned here: https://github.com/scipp/scippneutron/issues/458 
we might not have to implement them separately in nmx package.

### DIFF
--- a/src/ess/nmx/__init__.py
+++ b/src/ess/nmx/__init__.py
@@ -12,14 +12,14 @@ except importlib.metadata.PackageNotFoundError:
 del importlib
 
 from .data import small_mcstas_sample
-from .mcstas_loader import InputFilepath, NMXData, load_mcstas_nexus
+from .mcstas_loader import InputFilepath, NMXData, load_mcstas2_nexus
 from .reduction import NMXReducedData, TimeBinSteps
 
 __all__ = [
     "small_mcstas_sample",
     "NMXData",
     "InputFilepath",
-    "load_mcstas_nexus",
+    "load_mcstas2_nexus",
     "NMXReducedData",
     "TimeBinSteps",
 ]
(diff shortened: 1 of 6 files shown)


src/ess/nmx/__init__.py
```diff
@@ -12,14 +12,14 @@
...
-from .mcstas_loader import InputFilepath, NMXData, load_mcstas_nexus
+from .mcstas_loader import InputFilepath, NMXData, load_mcstas2_nexus
 from .reduction import NMXReducedData, TimeBinSteps
 
 __all__ = [
     "small_mcstas_sample",
     "NMXData",
     "InputFilepath",
-    "load_mcstas_nexus",
+    "load_mcstas2_nexus",
```
SimonHeybrock, line 22: Will we not be able to auto-detect the version of the file? It seems either error-prone or tedious, if the user has to know and specify the file version?

YooSunYoung: I couldn't find an easy way to auto-detect the version.
I asked Peter as well and he couldn't find the version information in the file either.

YooSunYoung: So according to Peter, if we can't find a proper version name in the ``creator`` attribute, 
it most-likely means that it's verison `2.*`/`(<3)`.
So... we can try parsing version number and use `mcstas2` loader if it fails...?

YooSunYoung:
I've received mcstas 3.4 files so I'll look into it and come up with an auto-detect solution as Simon suggested.

PR #32, opened 2024-02-15
Update dependencies.

YooSunYoung:
To use latest sciline.


YooSunYoung:
Can't update this until the next release since ``Callable`` caused error in ``sciline.sphinxext``.

PR #33, opened 2024-02-15
McStas loader to support McStas 3 results.

### LINKED ISSUE #31: McStas 2.7 and 3.4 version auto-detect in the loader.
The files are almost same, but there are some differences that we should handle.

|   | McStas 2.7 | McStas 3.4 |
|---|---|---|
| Event Data | All in one bank  |  `N` banks |
| Geometry | May have to strip some blanks to parse numbers | Tidy format, hence parsing may be easier |
| Version Info | May not be available, or should be available in the root attrs, ``creator`` | Should be available in the root attrs, ``creator`` |

Geometry files look different due to the tidier format,
but the contents and the structure are the same, so we can keep the 2.7 version for 3.4.

Here are the example geometry xml extracted from sample files: 
[mcstas_geometry.zip](https://github.com/scipp/essnmx/files/14277524/mcstas_geometry.zip)

TODO:
- [ ]  fix #26 together ~Couldn't fix it after all...!~ We can fix it now..!

YooSunYoung:
Fixes #31 

It didn't have to detect the version after all.
It's still not using detector name to retrieve data banks.

### DIFF
--- a/docs/examples/workflow.ipynb
+++ b/docs/examples/workflow.ipynb
@@ -6,7 +6,7 @@
    "source": [
     "# Workflow\n",
     "\n",
-    "## Collect Parameters and Providers\n",
+    "## Build Pipeline: Collect Parameters and Providers\n",
     "### Simulation(McStas) Data\n",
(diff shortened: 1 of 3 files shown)


YooSunYoung:
Closing this in favor of #35

PR #35, opened 2024-02-16
Fix McStas loader to support McStas 3 results.

### LINKED ISSUE #26: Only allow keyword arguments in ``load_mcstas_nexus``.
Loader accepts many arguments, so it'd better accept keyword arguments only.
Once sciline is released with the keyword arguments feature,
we can turn the loader into keyword-only function.

### LINKED ISSUE #31: McStas 2.7 and 3.4 version auto-detect in the loader.

YooSunYoung:
Continuing from #33 

Fixes #31 
Fixes #26 

## Main Changes
- Assuming multiple banks in dataset
- Allow optional rotation matrix for sample geometry
- Allow selecting detector name to compose coordinate from geometry


src/ess/nmx/mcstas_loader.py
```diff
@@ -29,26 +31,33 @@
...
 
-def _retrieve_event_list_name(keys: Iterable[str]) -> str:
-    prefix = "bank01_events_dat_list"
+def _retrieve_event_list_names(keys: Iterable[str]) -> tuple[str, ...]:
+    import re
 
-    # (weight, x, y, n, pixel id, time of arrival)
     mandatory_fields = 'p_x_y_n_id_t'
+    # (weight, x, y, n, pixel id, time of arrival)
+    pattern = r"bank(\d+\d+)_events_dat_list_" + mandatory_fields
```
jl-wynen, line 39:
```suggestion
    pattern = r"bank(\d\d+)_events_dat_list_" + mandatory_fields
```
Is this what you mean? 2 or more digits?

src/ess/nmx/mcstas_loader.py
```diff
@@ -182,8 +193,12 @@ def load_mcstas_nexus(
 
     from .mcstas_xml import read_mcstas_geometry_xml
 
+    # with snx.File(file_path) as file:
+    # mcstas_version = _retrieve_mcstas_version(file)
```
jl-wynen, line 197: Remove?

jl-wynen (commented):
Are there tests for this?

src/ess/nmx/mcstas_loader.py
YooSunYoung: Aaaah yes! thanks : D

YooSunYoung:
> Are there tests for this?

Yes...! Here: https://github.com/scipp/essnmx/blob/mcstas3-loader/tests/loader_test.py

jl-wynen:
> > Are there tests for this?
> 
> Yes...! Here: https://github.com/scipp/essnmx/blob/mcstas3-loader/tests/loader_test.py

Do they test both McStas 2 and 3 files?

YooSunYoung:
Here is the another script to make a subset of McStas 3 file for testing.
It is slightly different from the previous one since it now has 3 data banks.

```python
from pathlib import Path
import h5py
import os

file_path =  Path('mccode.h5')
copied_file_path = Path('mccode.h5'.replace('.h5', '_subset.h5'))

print(f"Copy subset of {file_path} to {copied_file_path}")

# Remove file path if exists.
if os.path.exists(copied_file_path):
    os.remove(copied_file_path)

# Copy all fields and subset of event data field.
# If you copy a file and remove a field, the file size does not change.
with h5py.File(file_path, 'r') as file:
    entry = file['entry1']
    dataset = entry['data']
    event_paths = [f'bank0{i}_events_dat_list_p_x_y_n_id_t' for i in range(1, 4)]
    event_map = {
        event_path: dataset[event_path]['events'][()] for event_path in event_paths
    }

    print("Original Shape: ", [events.shape for events in event_map.values()])

    with h5py.File(copied_file_path, 'w') as copied_file:
        copied_entry = copied_file.create_group('entry1')
        copied_data = copied_entry.create_group('data')
        
        # Copy all non-data fields.
        for key in entry.keys():
            if key == 'data':
               continue
            else:
                entry.copy(key, copied_entry, key)
        
        # Copy all non-event data fields.
        for data_key in copied_entry['data'].keys():
            if data_key not in event_paths:
                dataset.copy(data_key, copied_data, data_key)

        # Copy subset of events.
        for event_path, events in event_map.items():
            copied_data.create_group(event_path)
            copied_data[event_path].create_dataset('events', data=events)
            print(event_path, " subset Shape: ", events.shape)

```

src/ess/nmx/data/__init__.py
```diff
@@ -15,15 +15,38 @@ def _make_pooch():
...
-def small_mcstas_sample():
-    return get_path('small_mcstas_sample.h5')
+def small_mcstas_2_sample():
+    """McStas 2 file containing small number of events."""
+    import warnings
+
+    warnings.warn(
+        DeprecationWarning(
+            '``essnmx`` will not support loading files '
+            'made by McStas with version less than 3 from ``25.0.0``. '
```
YooSunYoung, line 35: @Justin-Bergmann 

Does it make sense to drop support for McStas 2 at some point, i.e. from `2025`?

YooSunYoung:
> > > Are there tests for this?
> > 
> > 
> > Yes...! Here: https://github.com/scipp/essnmx/blob/mcstas3-loader/tests/loader_test.py
> 
> Do they test both McStas 2 and 3 files?

I added tests with McStas 3 file as well...!

PR #51, opened 2024-03-15
fix: load single bank

### LINKED ISSUE #49: Handle banks via workflow graph instead of via extra dimension
The current McStas loader loads all banks and concats the events,
https://github.com/scipp/essnmx/blob/c5a2520e1ac6b209c855552729c6b0076abf4453/src/ess/nmx/mcstas_loader.py#L50-L60
subsequently folding after grouping by pixel
https://github.com/scipp/essnmx/blob/c5a2520e1ac6b209c855552729c6b0076abf4453/src/ess/nmx/mcstas_loader.py#L134

- Change this so only a single bank is handled in the code (loading, reduction steps).
- Multiple banks should be handled via Sciline, i.e., in the workflow graph. This will, e.g., allow for processing only a single bank at a time, saving memory.
  - We can use a param table to define the detector banks that should be processed.
- Saving to NeXus may need to change slightly. Need to check with the IDS.
  - I do not understand the current code, the `panel` dimension is first removed in https://github.com/scipp/essnmx/blob/c5a2520e1ac6b209c855552729c6b0076abf4453/src/ess/nmx/reduction.py#L245-L247 and then a panel dim of length 1 is added back into the dataset when saving, in https://github.com/scipp/essnmx/blob/c5a2520e1ac6b209c855552729c6b0076abf4453/src/ess/nmx/reduction.py#L151-L153.

jokasimr:
Fixes #49

### DIFF
--- a/docs/examples/workflow.ipynb
+++ b/docs/examples/workflow.ipynb
@@ -31,6 +31,7 @@
     "    event_weights_from_probability,\n",
     "    ProtonChargeConverter,\n",
     "    proton_charge_from_event_data,\n",
+    "    DetectorBankName,\n",
     ")\n",
     "from ess.nmx.data import small_mcstas_3_sample\n",
     "from ess.nmx.reduction import bin_time_of_arrival, TimeBinSteps\n",
@@ -45,6 +46,7 @@
     "    MaximumProbability: DefaultMaximumProbability,\n",
     "    EventWeightsConverter: event_weights_from_probability,\n",
     "    ProtonChargeConverter: proton_charge_from_event_data,\n",
+    "    DetectorBankName: 'bank01',\n",
     "}"
    ]
   },
--- a/docs/examples/workflow.ipynb
+++ b/docs/examples/workflow.ipynb
@@ -8,60 +8,9 @@
     "In this example, we will use McStas 3 simulation file.\n",
     "\n",
     "## Build Pipeline (Collect Parameters and Providers)\n",
-    "There is a dedicated loader, ``load_mcstas_nexus`` for ``McStas`` simulation data workflow. <br>\n",
-    "``MaximumProbability`` can be manually provided to the loader <br>\n",
-    "to derive more realistic number of events. <br>\n",
+    "Import the providers from ``load_mcstas_nexus`` to use the ``McStas`` simulation data workflow. <br>\n",
+    "``MaximumProbability`` can be manually provided to derive more realistic number of events. <br>\n",
     "It is because ``weights`` are given as probability, not number of events in a McStas file. <br>"
-   ]
-  },
-  {
-   "cell_type": "code",
-   "execution_count": null,
-   "metadata": {},
-   "outputs": [],
-   "source": [
-    "# Collect parameters and providers\n",
-    "import scipp as sc\n",
-    "from ess.nmx.mcstas_loader import load_mcstas_nexus\n",
-    "from ess.nmx.mcstas_loader import (\n",
-    "    InputFilepath,\n",
-    "    MaximumProbability,\n",
-    "    DefaultMaximumProbability,\n",
-    "    EventWeightsConverter,\n",
-    "    event_weights_from_probability,\n",
-    "    ProtonChargeConverter,\n",
-    "    proton_charge_from_event_data,\n",
-    "    DetectorBankName,\n",
-    ")\n",
-    "from ess.nmx.data import small_mcstas_3_sample\n",
-    "from ess.nmx.reduction import bin_time_of_arrival, TimeBinSteps\n",
-    "\n",
-    "providers = (load_mcstas_nexus, bin_time_of_arrival, )\n",
-    "\n",
-    "file_path = small_mcstas_3_sample()  # Replace it with your data file path\n",
-    "params = {\n",
-    "    TimeBinSteps: TimeBinSteps(50),\n",
-    "    InputFilepath: InputFilepath(file_path),\n",
-    "    # Additional parameters for McStas data handling.\n",
-    "    MaximumProbability: DefaultMaximumProbability,\n",
-    "    EventWeightsConverter: event_weights_from_probability,\n",
-    "    ProtonChargeConverter: proton_charge_from_event_data,\n",
-    "    DetectorBankName: 'bank01',\n",
-    "}"
-   ]
-  },
-  {
-   "cell_type": "markdown",
-   "metadata": {},
-   "source": [
-    "``event weights converter`` and ``proton_charge_from_event_data`` are\n",
-    "\n",
-    "set as parameters for reproducibility of workflow and accessibility to the documentation.\n",
-    "\n",
-    "The reason of having them as parameters not as providers is,\n",
-    "\n",
-    "1. They are not part of general reduction, which are only for McStas cases.\n",
-    "2. They are better done while the file is open and read in the loader."
    ]
   },
   {
@@ -72,8 +21,37 @@
    "source": [
     "import sciline as sl\n",
     "\n",
-    "nmx_pl = sl.Pipeline(list(providers), params=params)\n",
-    "nmx_pl"
+    "from ess.nmx.mcstas_loader import providers as loader_providers\n",
+    "from ess.nmx.mcstas_xml import read_mcstas_geometry_xml\n",
+    "from ess.nmx.data import small_mcstas_3_sample, small_mcstas_2_sample\n",
+    "from ess.nmx.types import *\n",
+    "from ess.nmx.reduction import NMXData, NMXReducedData, bin_time_of_arrival, TimeBinSteps\n",
+    "\n",
+    "\n",
+    "pl = sl.Pipeline((*loader_providers, read_mcstas_geometry_xml, bin_time_of_arrival))\n",
+    "# Replace with the path to your own file\n",
+    "pl[FilePath] = small_mcstas_3_sample()\n",
+    "pl[MaximumProbability] = 10000\n",
+    "pl[TimeBinSteps] = 50\n",
+    "# DetectorIndex selects what detector panels to include in the run\n",
+    "# in this case we select all three panels.\n",
+    "pl.set_param_table(sl.ParamTable(RunID, {DetectorIndex: range(3)}, index=range(3)))"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "To see what the pipeline can produce, display it:"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "pl"
    ]
   },
   {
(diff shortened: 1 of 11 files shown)


src/ess/nmx/mcstas_loader.py
```diff
@@ -123,15 +102,10 @@ def _compose_event_data_array(
...
-
-    num_panels:
-        The number of (detector) panels used in the experiment.
-
     """
 
     events = sc.DataArray(data=weights, coords={'t': t_list, 'id': id_list})
-    grouped: sc.DataArray = events.group(pixel_ids)
-    return grouped.fold(dim='id', sizes={'panel': num_panels, 'id': -1})
+    return events.group(pixel_ids).fold(dim='id', sizes={'panel': 1, 'id': -1})
```
SimonHeybrock, line 108: Shouldn't the `panel` dimension be gone now? Why do we still have to fold?

src/ess/nmx/mcstas_loader.py
```diff
@@ -1,6 +1,6 @@
 # SPDX-License-Identifier: BSD-3-Clause
 # Copyright (c) 2023 Scipp contributors (https://github.com/scipp)
-from typing import Callable, Iterable, NewType, Optional
+from typing import Callable, NewType, Optional
 
 import scipp as sc
 import scippnexus as snx
```
SimonHeybrock, line 6: A couple lines below: Not your change, but why do we have `DetectorName` and `DetectorBankName`? Can we remove the latter?

src/ess/nmx/mcstas_loader.py
```diff
@@ -190,16 +165,24 @@ def load_mcstas_nexus(
...
 
     geometry = read_mcstas_geometry_xml(file_path)
-    detectors = [det.name for det in geometry.detectors]
-    coords = geometry.to_coords(*detectors)
+    bank_name_to_detector_name = dict(
+        zip(
+            (f'bank0{i}' for i in range(1, 4)),
+            ('nD_Mantid_0', 'nD_Mantid_1', 'nD_Mantid_2'),
+        )
+    )
```
SimonHeybrock, line 181: It seems like the `read_mcstas_geometry_xml` is loading all banks? Is this something that should also be split up, alongside these changes, instead of disentangling it here? Hard-coding the detector names here looks very odd anyway.

src/ess/nmx/mcstas_loader.py
jokasimr: For now just because some of the tests expects it to be there. I'll remove it and change the tests.

src/ess/nmx/mcstas_loader.py
jokasimr: Yeah should have put a comment that this is a placeholder solution. I don't really know how to to this disentangling in a sensible way. It's possible to do it based on the McStas run definition code, but that seems a bit cumbersome.

jokasimr: I'm not sure what you mean. The old code loaded all detectors and here we only want to load one.

SimonHeybrock: The old code did not have to spell out the mapping here. Why do we have to now?

jokasimr: Because we only want to load one bank, and also only the detector that is associated with the bank we're loading. The old code loaded all banks and also loaded all detectors.

SimonHeybrock: My points is: The previous implementation simply used a positional (sorted) association. There was no need to hard-code a mapping here, relying on any particular naming scheme. Why can't we keep doing the same, e.g., by using a bank index?

jokasimr: Yeah maybe we can do that. I mean, that depends on if we can rely on the ordering of the banks being the same as the ordering of the detectors. The order of the banks is the order they occur in the nexus file I guess, and the entries in a HDF group are ordered alphanumerically.

The old code doesn't really need the banks and the detectors having the same order (I think) it only assumes the detectors are ordered by their respective `detector_ids`.

src/ess/nmx/mcstas_loader.py
```diff
@@ -190,16 +165,24 @@ def load_mcstas_nexus(
...
 
     from .mcstas_xml import read_mcstas_geometry_xml
 
     geometry = read_mcstas_geometry_xml(file_path)
-    detectors = [det.name for det in geometry.detectors]
-    coords = geometry.to_coords(*detectors)
+    bank_name_to_detector_name = dict(
+        zip(
+            (f'bank0{i}' for i in range(1, 4)),
+            ('nD_Mantid_0', 'nD_Mantid_1', 'nD_Mantid_2'),
```
YooSunYoung, line 179: I think we don't really need this dictionary if we use an integer index of detector/bank, and make sure we sorted them in the order of pixel ids they carry as Simon mentioned.

YooSunYoung (commented):
I think it'll be easier to use integer index of detector instead of hard-coded detector name.
We just need to make sure that they are sorted in the order of pixel-ids.

And if there is no events to compare the pixel ids in the banks,
it doesn't matter really which detector bank to use, so it can be just sorted by their names
Even though there is no events, it should still be possible to load the geometry, for example ``pixel-ids`` and ``pixel-offsets``.

jokasimr:
Noticed just now that one of the two test files (mcstas2) puts all events from the three detectors into the same event data list.
It makes things complicated because in that case things needs to work differently from the case when each event data bank is associated with a separate detector.

I'm wondering if this is something we should support right now, I think it would be easier to develop this from the assumption that each `event_data_bank` is associated with only one detector panel.

YooSunYoung:
> Noticed just now that one of the two test files (mcstas2) puts all events from the three detectors into the same event data list.

Yeah... that was why it was loading in all possible banks from the first place.
I think we can just use the only bank if it fails to find the related one.
It'll drop all unused pixels anyways.

> I'm wondering if this is something we should support right now, I think it would be easier to develop this from the assumption that each event_data_bank is associated with only one detector panel.

I'm not sure ... especially since Aaron is in process of handling over Justin's work these days.
@Justin-Bergmann , @aaronfinke do you still need McStas 2 results?

src/ess/nmx/mcstas_loader.py
jokasimr: Replaced the hardcoded association with a regex that looks in the mcstas definition file for detectors and the event datasets they are associated with.

docs/examples/workflow.ipynb
```diff
@@ -31,6 +31,7 @@
     "    event_weights_from_probability,\n",
     "    ProtonChargeConverter,\n",
     "    proton_charge_from_event_data,\n",
+    "    DetectorBankName,\n",
```
YooSunYoung, line 34: Now that we can map the `detector name` to `detector bank name`,
should we maybe use `detector name` instead of the bank name as a user input...?
Which one comes first...? `DetectorName` or `DetectorBankName`...?

src/ess/nmx/mcstas_loader.py
```diff
@@ -31,33 +32,13 @@
...
-        return re.search(pattern, key) is not None
-
-    if not (matching_keys := tuple(filter(_filter_event_list_name, keys))):
-        raise ValueError("Can not find event list name.")
-
-    return matching_keys
-
-
-def _retrieve_raw_event_data(file: snx.File) -> sc.Variable:
+def _retrieve_raw_event_data(file: snx.File, bank_name: str) -> sc.Variable:
```
YooSunYoung, line 35: I thought we could find the full name of the detector data bank in the mcstas description.
Is it just prefix that we can find...?

docs/examples/workflow.ipynb
```diff
@@ -141,6 +115,15 @@
...
   },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "binned_dg.export_as_nexus('test.nxs')"
+   ]
+  },
```
YooSunYoung, line 126: Can you remove this code cell? It might make documentation build quite slow.

src/ess/nmx/const.py
```diff
@@ -0,0 +1,5 @@
+DETECTOR_DIM = 'panel'
+DETECTOR_SHAPE = (1280, 1280)
+PIXEL_DIM = 'id'
+TOF_DIM = 't'
+DETECTOR_DIM = 'panel'
```
YooSunYoung, line 5: This line seems same as line #1.

src/ess/nmx/mcstas_loader.py
```diff
@@ -1,140 +1,103 @@
...
-    import re
 
-    mandatory_fields = 'p_x_y_n_id_t'
-    # (weight, x, y, n, pixel id, time of arrival)
-    pattern = r"bank(\d\d+)_events_dat_list_" + mandatory_fields
 
-    def _filter_event_list_name(key: str) -> bool:
-        return re.search(pattern, key) is not None
+def detector_name_from_index(index: DetectorIndex) -> DetectorName:
+    return f'nD_Mantid_{index}'
```
YooSunYoung, line 26: Is this function not used?

src/ess/nmx/mcstas_xml.py
```diff
@@ -384,26 +384,25 @@ def to_coords(self, *det_names: str) -> dict[str, sc.Variable]:
...
 
-        return {
+        coords = {
             'pixel_id': _construct_pixel_ids(detectors),
-            'fast_axis': sc.concat(fast_axes, detector_dim),
-            'slow_axis': sc.concat(slow_axes, detector_dim),
-            'origin_position': sc.concat(origins, detector_dim),
+            'fast_axis': sc.concat(fast_axes, DETECTOR_DIM),
+            'slow_axis': sc.concat(slow_axes, DETECTOR_DIM),
+            'origin_position': sc.concat(origins, DETECTOR_DIM),
```
YooSunYoung, line 396: Will they still have a `DETECTOR_DIM` even if it is only from 1 detector...?
If yes, should `pixel_id` and `position` also have `DETECTOR_DIM`...?

src/ess/nmx/types.py
```diff
@@ -0,0 +1,36 @@
...
+
+ProtonCharge = NewType("ProtonCharge", sc.Variable)
+"""The proton charge signal"""
+
+CrystalRotation = NewType("CrystalRotation", sc.Variable)
+"""Rotation of the crystal"""
+
+RunID = NewType("RunId", int)
+
+DetectorGeometry = NewType("DetectorGeometry", Any)
```
YooSunYoung, line 36: Will you also update the api-reference page to have ``nmx.types``...?

YooSunYoung (changes requested):

src/ess/nmx/const.py
jokasimr: I don't understand what you mean ;) does this code already exist elsewhere?

src/ess/nmx/mcstas_loader.py
jokasimr: Yes it's used, it's exported as a provider.

src/ess/nmx/mcstas_xml.py
jokasimr: Yes they will still have the detector dimension even if it is only one detector.
I'm not sure if they should or not. But I figured that it doesn't really matter if they do, so I left it.

docs/examples/workflow.ipynb
jokasimr: Yes sure, good point.

src/ess/nmx/const.py
YooSunYoung: Ah sorry I meant line#5 and line#1 are repeated.

src/ess/nmx/mcstas_loader.py
jokasimr: We can only get the prefix, the full name is something that contains all the variables in the data: `bank01_events_dat_list_p_x_y_n_id_t` but the name we extract form the mcstas description is just `bank_01_events.dat`

YooSunYoung (approved):

PR #54, opened 2024-04-12
MTZ IO for scaling 

YooSunYoung:
related to #48 

Providers for MTZ file I/O and data frame reducing.

Reducing steps are divided to 2 steps since part of the steps should be done on an individual dataset and part of the steps should be done on the merged dataset.
Rest of the procedures will be in the next PRs.


src/ess/nmx/mtz_io.py
```diff
@@ -0,0 +1,202 @@
...
+import numpy as np
+import pandas as pd
+import sciline as sl
+
+# Index types for param table.
+MTZFileIndex = NewType("MTZFileIndex", int)
+"""The index of the mtz file when iterating over multiple mtz files."""
+
+# User defined or configurable types
+MTZFilepath = NewType("MTZFilepath", str)
```
jl-wynen, line 15: Should be a `Path`: https://scipp.github.io/essreduce/user-guide/reduction-workflow-guidelines.html#c-1-use-common-names-and-types

src/ess/nmx/mtz_io.py
```diff
@@ -0,0 +1,202 @@
...
+"""Path to the mtz file"""
+LambdaBinSize = NewType("LambdaBinSize", int)
```
jl-wynen, line 17:
```suggestion
LambdaBinCount = NewType("LambdaBinCount", int)
```
?
`Size` makes me think it is the bind width.

src/ess/nmx/mtz_io.py
```diff
@@ -0,0 +1,202 @@
...
+"""The number of bins to use for binning wavelength(lambda)."""
+SpaceGroupDesc = NewType("SpaceGroupDesc", str)
+"""The space group description. e.g. 'P 21 21 21'"""
+DEFAULT_SPACE_GROUP_DESC = SpaceGroupDesc("P 21 21 21")
+"""The default space group description to use if not found in the mtz files."""
+
+# Computed types
+RawMtz = NewType("RawMtz", gemmi.Mtz)
+"""The mtz file as a gemmi object"""
+RawMtzDataFrame = NewType("RawMtzDataFrame", pd.DataFrame)
```
jl-wynen, line 27: Without having looked in detail yet, why does this use pandas? Can we use scipp data arrays or datasets instead?

jl-wynen (commented):
Only a first pass. Please see my question about pandas

src/ess/nmx/mtz_io.py
YooSunYoung: Actually it's not used in this scope so I removed it for now. I'll apply it in along with the relevant changes. Thank you!

src/ess/nmx/mtz_io.py
YooSunYoung: I guess...? It just has a helper to convert it into pandas data frame and it was easier to use pandas for flexible modification on a column.

We need scipp for later process of binning though.

SimonHeybrock: I think using Pandas is fine in general. Not everything has to be done with Scipp, if Pandas is more convenient. If the data (columns) have units though, then we should favor Scipp.

src/ess/nmx/mtz_io.py
```diff
@@ -0,0 +1,201 @@
...
+
+    # HKL should always be integer.
+    mtz_df[["H", "K", "L"]] = orig_df[["H", "K", "L"]].astype(int)
+    mtz_df["hkl"] = mtz_df[["H", "K", "L"]].values.tolist()
+
+    def _calculate_d(row: pd.Series) -> float:
+        return mtz.get_cell().calculate_d(row["hkl"])
+
+    mtz_df["d"] = mtz_df.apply(_calculate_d, axis=1)
+    # $(2d)^{-2} = \sin^2(\theta)/\lambda^2
```
jl-wynen, line 103:
```suggestion
    # (2d)^{-2} = \sin^2(\theta)/\lambda^2
```

src/ess/nmx/mtz_io.py
```diff
@@ -0,0 +1,201 @@
...
+    Spacegroup is always expected in any MTZ files, but it may be missing.
+
+    Parameters
+    ----------
+    mtzs:
+        A series of raw mtz datasets.
+
+    spacegroup_desc:
+        The space group description to use if not found in the mtz files.
+        If None, ``DEFAULT_SPACE_GROUP_DESC`` is used.
```
jl-wynen, line 128: Can you refer to the actual object? I.e.
```
:attr:`DEFAULT_SPACE_GROUP_DESC`
```
might work if the docs actually show attrs.

YooSunYoung: Yeah it was
```python
:attr:`~DEFAULT_SPACE_GROUP_DESC`
```
Thanks!

jl-wynen (approved):

PR #55, opened 2024-04-15
Deriving scale factor from reference bin.

YooSunYoung:
Related to #48


YooSunYoung:
@jl-wynen 
It's still in-progress but can you please review this one as well...?

src/ess/nmx/mtz_io.py
```diff
@@ -20,6 +21,12 @@
 DEFAULT_SPACE_GROUP_DESC = SpaceGroupDesc("P 21 21 21")
 """The default space group description to use if not found in the mtz files."""
 
+# Custom column names
+WavelengthColumnName = NewType("WavelengthColumnName", str)
+"""The name of the wavelength column in the mtz file."""
+DEFUAULT_WAVELENGTH_COLUMN_NAME = WavelengthColumnName("LAMBDA")
```
jl-wynen, line 27:
```suggestion
DEFAULT_WAVELENGTH_COLUMN_NAME = WavelengthColumnName("LAMBDA")
```

src/ess/nmx/mtz_io.py
```diff
@@ -186,15 +203,70 @@ def _rapio_asu_to_asu(row: pd.Series) -> list[int]:
 
     merged_df["hkl_eq"] = merged_df.apply(_rapio_asu_to_asu, axis=1)
 
+    def unpack_vector(row: pd.Series, *new_names) -> pd.DataFrame:
+        return pd.DataFrame(
+            {name: [val[i] for val in row] for i, name in enumerate(new_names)}
+        )
```
jl-wynen, line 209: There must be a better way to do this directly with a dataframe. I don't know how many rows there can be, but indexing like this and building intermediate lists can get quite expensive.

I'm also not sure what this accomplishes. Doesn't this simply rename columns?

src/ess/nmx/mtz_io.py
```diff
@@ -186,15 +203,70 @@ def _rapio_asu_to_asu(row: pd.Series) -> list[int]:
...
+    # Convert to scipp Dataset
+    nmx_mtz_ds = from_pandas_dataframe(
+        to_scipp,
+        data_columns=dummy_data_column_name,
+        header_parser=parse_bracket_header,
+    )
+    # Add back the vector columns
+    for col, values in vector_coords.items():
+        nmx_mtz_ds.coords[col] = sc.vectors(
+            dims=nmx_mtz_ds.dims, values=[val for val in values]
```
jl-wynen, line 248: Can't you convert `values` to a numpy array directly?

src/ess/nmx/mtz_io.py
```diff
@@ -186,15 +203,70 @@ def _rapio_asu_to_asu(row: pd.Series) -> list[int]:
...
 
+def nmx_mtz_dataframe_to_scipp_dataarray(
+    nmx_mtz_df: NMXMtzDataFrame,
+) -> NMXMtzDataArray:
+    """Converts the reduced mtz dataframe to a scipp dataarray."""
+    from scipp.compat.pandas_compat import from_pandas_dataframe, parse_bracket_header
+
+    # Add unit to the name
+    to_scipp = nmx_mtz_df.copy(deep=False)
+    to_scipp[DEFUAULT_WAVELENGTH_COLUMN_NAME + " [Å]"] = to_scipp[
```
jl-wynen, line 227: Seems like a hack. You could also assign the units to the final dataset directly.

src/ess/nmx/scaling.py
```diff
@@ -0,0 +1,91 @@
...
+    while 0 < (cur_idx := middle_number + offset) < len(binned) and _is_bin_empty(
+        binned, cur_idx
+    ):
+        offset = -offset + 1 if offset <= 0 else -offset
+
+    if _is_bin_empty(binned, cur_idx):
+        raise ValueError("No reference group found.")
+
+    ref: sc.DataArray = binned[cur_idx].values.copy(deep=False)
+    grouped: sc.DataArray = ref.group("hkl_eq_hash")
```
jl-wynen, line 78: I think you should create the hashed coord here instead of in the pandas to scipp converter because it is a detail of how you group here.

And then, it should be removed after grouping and the event `"hkl_eq"` coord should be lifted to a bin coord. (Unless you need to regroup later and still require the hashed coord.)

src/ess/nmx/mtz_io.py
```diff
@@ -186,15 +203,70 @@ def _rapio_asu_to_asu(row: pd.Series) -> list[int]:
...
+    )
+    # Add back the vector columns
+    for col, values in vector_coords.items():
+        nmx_mtz_ds.coords[col] = sc.vectors(
+            dims=nmx_mtz_ds.dims, values=[val for val in values]
+        )
+    # Add HKL EQ hash coordinate for grouping
+    nmx_mtz_ds.coords["hkl_eq_hash"] = sc.Variable(
+        dims=nmx_mtz_ds.dims,
+        values=[hash(tuple(val)) for val in nmx_mtz_ds.coords["hkl_eq"].values],
```
jl-wynen, line 253: I don't know how likely hash collisions are for tuples of numbers. But do you know the possible ranges of `h`, `k`, `l`? If so, you could merge them using `h * nh * nk + k * nk + l` and guarantee that there are no collisions.

src/ess/nmx/mtz_io.py
YooSunYoung: If I use `to_numpy()`, it throws this error
```bash
ValueError: Data has too few dimensions for given dimension labels.
```

src/ess/nmx/mtz_io.py
YooSunYoung: It's when we want to use `H`, `K`, `L` separately. 
It was used in the next steps, but I changed that part so maybe we can skip this here. I'll get rid of it.

src/ess/nmx/mtz_io.py
YooSunYoung: Oh... right... I didn't think about the hash collisions.. 
I'll hash string instead then...

src/ess/nmx/mtz_io.py
jl-wynen: Strange. Does `to_numpy` return a flattened array? Or an array of objects?

src/ess/nmx/scaling.py
```diff
@@ -37,14 +37,42 @@ def get_lambda_binned(
...
+        dims=var.dims,
+        values=apply_func(var.values),
+    )
+
+
+def hash_variable(var: sc.Variable) -> sc.Variable:
+    """Hash the coordinate values."""
+
+    def _hash_repr(val: Any) -> int:
+        return hash(str(val))
```
jl-wynen, line 71: This doesn't improve anything. Any hash can lead to collisions. This is why I asked about a specific way of getting a unique combination of the 3 numbers. But if you don't know the ranges of hkl, you can't to that, AFAIK.
Have you tried using `str(val)` instead of `hash(str(val))`?

YooSunYoung: Yes...! I updated how it makes a unique key. Now it's not using hash.

src/ess/nmx/mtz_io.py
YooSunYoung: It looks like an array of objects.

YooSunYoung:
#57  Also updates the helper functions in  `reduction.py`,
so I would rather update it there instead of here...! Thanks!

src/ess/nmx/reduction.py
```diff
@@ -243,3 +244,124 @@ def bin_time_of_arrival(
...
     )
+
+
+def _apply_elem_wise(
+    func: Callable, var: sc.Variable, *, result_dtype: Any = None
+) -> sc.Variable:
+    """Apply a function element-wise to the variable values.
+
+    This helper is only for vector-dtype variables.
+    Use ``numpy.vectorize`` for other types.
```
SimonHeybrock, line 255: Looks inefficient, why is this needed?

YooSunYoung:
I think the `I` column should just be the data and we don't need `dummy_data`. 
I'll reopen it agian after fixing it...!

YooSunYoung:
So I fixed the data and the grouping methods.
It's still a detouring grouping.
This requirement was requested for a while ago,
so I would like to prioritize delivering this
and we can fix the implementation detail later as mentioned in the comments.

src/ess/nmx/mtz_io.py
```diff
@@ -64,14 +79,20 @@ def mtz_to_pandas(mtz: gemmi.Mtz) -> pd.DataFrame:
     )
 
 
-def reduce_single_mtz(mtz: RawMtz) -> RawMtzDataFrame:
+def reduce_single_mtz(
+    mtz: RawMtz,
+    lambda_column_name: WavelengthColumnName = DEFAULT_WAVELENGTH_COLUMN_NAME,
```
SimonHeybrock, line 84: Note that Sciline never uses default values.

src/ess/nmx/mtz_io.py
```diff
@@ -185,16 +209,69 @@ def _rapio_asu_to_asu(row: pd.Series) -> list[int]:
...
+            DEFAULT_SIGMA_INTENSITY_COLUMN_NAME,
+        ],
+        header_parser=parse_bracket_header,
+    )
+    # Add units
+    nmx_mtz_ds.coords[DEFAULT_WAVELENGTH_COLUMN_NAME].unit = sc.units.angstrom
+    for key in nmx_mtz_ds.keys():
+        nmx_mtz_ds[key].unit = sc.units.dimensionless
+    for coord in ("H", "K", "L", "H_EQ", "K_EQ", "L_EQ"):
+        nmx_mtz_ds.coords[coord].unit = sc.units.dimensionless
```
SimonHeybrock, line 254: Dimensionless or `None`?

src/ess/nmx/reduction.py
```diff
@@ -243,3 +244,91 @@ def bin_time_of_arrival(
...
+        raise ValueError("All columns must be integer type.")
+    # Check if all columns have the same dimensions
+    dims = set(var.dim for var in vars)
+    if len(dims) != 1:
+        raise ValueError("All columns must have the same dimensions.")
+    return sc.array(
+        dims=dims,
+        values=[
+            splitter.join(str(val) for val in row)
+            for row in zip(*(var.values for var in vars))
```
SimonHeybrock, line 261: Goes wrong if lengths differ?

src/ess/nmx/reduction.py
```diff
@@ -243,3 +244,91 @@ def bin_time_of_arrival(
...
+    separated = [val.split(splitter) for val in var.values]
+    # Check if all rows have the same length
+    lengths = set(len(row) for row in separated)
+    if len(lengths) != 1:
+        raise ValueError("All rows must have the same length.")
+
+    return tuple(
+        sc.array(dims=var.dims, values=[int(row[i]) for row in separated], dtype=int)
+        for i in range(lengths.pop())
+    )
```
SimonHeybrock, line 278: I presume these variables of strings have very few elements?

src/ess/nmx/reduction.py
```diff
@@ -243,3 +244,91 @@ def bin_time_of_arrival(
...
+    )  # Must keep the order
+
+    tmp_coord = _join_variables(*group_coords)
+    copied.coords[tmp_str_coord_name] = tmp_coord
+
+    if all(isinstance(arg, str) for arg in args):
+        group_var = sc.array(
+            dims=[tmp_str_coord_name],
+            values=np.unique(copied.coords[tmp_str_coord_name].values),
+        )
```
SimonHeybrock, line 319: Is this a workaround for scipp/scipp#3425? If so, add a link in a comment.

src/ess/nmx/reduction.py
```diff
@@ -243,3 +244,91 @@ def bin_time_of_arrival(
...
+        raise ValueError("All coordinates must be either str or sc.Variable.")
+
+    grouped = copied.group(group_var)
+    real_coords = _split_variable(grouped.coords[tmp_str_coord_name])
+    for i_coord, name in enumerate(group_coord_names):
+        grouped.coords[name] = real_coords[i_coord].rename_dims(
+            {real_coords[i_coord].dim: grouped.dim}
+        )
+
+    return grouped.drop_coords([tmp_str_coord_name])
```
SimonHeybrock, line 334: How many rows does this data have in practice, or more importantly, what are the ranges of H/K/L? If it is not too many, maybe we could do something simple like:
```python
grouped = da.group('H', 'K', 'L')
non_empty = grouped[grouped.bins.size() > 0]
return non_empty
```

src/ess/nmx/mtz_io.py
YooSunYoung: Yes, I am aware of it. But it makes tests and simple scripting easier with default values.

src/ess/nmx/reduction.py
YooSunYoung: It checks if they all have the same dimensions. I'll also add the lengths check though.

src/ess/nmx/reduction.py
YooSunYoung: You mean each string? There should only be 3, HKL.
And if you mean the the length of the `var`, it could be long ....
Hmm maybe we don't need to put the original coordinates back actually.

src/ess/nmx/reduction.py
SimonHeybrock: Forgot `flatten` in the suggestion, edited above.

src/ess/nmx/reduction.py
YooSunYoung: I did in the docstring, but I'll also add exactly which part is the workaround...

src/ess/nmx/reduction.py
YooSunYoung: I asked Aaron and Justin but the range varies from experiment to experiment,
and from the files I got, it looks like the ranges can be some hundreds.

We also need to group them only by the HKL groups that the reference bin has.
I can't think of an easy solution for that though... 
How would it be like ...?

src/ess/nmx/mtz_io.py
YooSunYoung: `None`...! Sorry I'll remove this part.

src/ess/nmx/reduction.py
YooSunYoung: Ah I missed the flatten. Hmm maybe...?

YooSunYoung: Ah but they are also binned by `wavelength`.
It drops the bin-edges after grouping....

YooSunYoung: I think for the reference group, it should be fine to use this method. I'll update it.

src/ess/nmx/reduction.py
SimonHeybrock: I mean the length of the variable.

src/ess/nmx/reduction.py
```diff
@@ -243,3 +244,76 @@ def bin_time_of_arrival(
...
+    return sc.array(
+        dims=dims,
+        values=[
+            splitter.join(str(val) for val in row)
+            for row in zip(*(var.values for var in vars))
+        ],
+    )
+
+
+def _zip_and_group(da: sc.DataArray, /, *args: str | sc.Variable) -> sc.DataArray:
```
SimonHeybrock, line 271: Where is this used now?

YooSunYoung: It was left for grouping other wavelength bins, but it's done in a different way after all.
I'll delete it.

src/ess/nmx/mtz_io.py
```diff
@@ -20,6 +21,19 @@
...
+"""The name of the intensity column in the mtz file."""
+DEFAULT_INTENSITY_COLUMN_NAME = IntensityColumnName("I")
+
+SigmaIntensityColumnName = NewType("SigmaIntensityColumnName", str)
+"""The name of the standard uncertainty of intensity column in the mtz file."""
+DEFAULT_SIGMA_INTENSITY_COLUMN_NAME = SigmaIntensityColumnName("SIGI")
+
 # Computed types
 RawMtz = NewType("RawMtz", gemmi.Mtz)
 """The mtz file as a gemmi object"""
```
SimonHeybrock, line 39: Some context about the MTZ files would be useful. In particular, what are the properties of the resulting `DataFrame`? How large is it, etc.?

src/ess/nmx/mtz_io.py
```diff
@@ -101,8 +122,11 @@ def _calculate_d(row: pd.Series) -> float:
     mtz_df["d"] = mtz_df.apply(_calculate_d, axis=1)
     # (2d)^{-2} = \sin^2(\theta)/\lambda^2
     mtz_df["resolution"] = (1 / mtz_df["d"]) ** 2 / 4
-
     mtz_df["I_div_SIGI"] = orig_df["I"] / orig_df["SIGI"]
+    mtz_df[DEFAULT_WAVELENGTH_COLUMN_NAME] = orig_df[lambda_column_name]
```
SimonHeybrock, line 126: Rename to wavelength in *output*, for consistency.

src/ess/nmx/scaling.py
```diff
@@ -0,0 +1,98 @@
...
+
+    """
+
+    return WavelengthBinned(
+        mtz_da.bin({DEFAULT_WAVELENGTH_COLUMN_NAME: wavelength_bin_size})
+    )
+
+
+def get_reference_bin(binned: WavelengthBinned) -> ReferenceWavelengthBin:
+    """Find the reference group in the binned dataset.
```
SimonHeybrock, line 51: Reference for what? -> Answer: Used for estimating the scale factor.

Returns the events of the reference bin.

`get_events_for_scale_factor_estimation`?

src/ess/nmx/scaling.py
```diff
@@ -0,0 +1,98 @@
...
+
+    if _is_bin_empty(binned, cur_idx):
+        raise ValueError("No reference group found.")
+
+    return binned[cur_idx].values.copy(deep=False)
+
+
+def calculate_scale_factor_per_hkl_eq(
+    ref_bin: ReferenceWavelengthBin,
+) -> ReferenceScaleFactor:
```
SimonHeybrock, line 82: Docstring? Missing equations. Rename.

SimonHeybrock (commented):
Notes from our chat:

src/ess/nmx/mtz_io.py
```diff
@@ -64,14 +79,20 @@ def mtz_to_pandas(mtz: gemmi.Mtz) -> pd.DataFrame:
     )
 
 
-def reduce_single_mtz(mtz: RawMtz) -> RawMtzDataFrame:
+def reduce_single_mtz(
```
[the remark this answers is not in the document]
YooSunYoung: It's not.. I renamed it...!

src/ess/nmx/mtz_io.py
YooSunYoung: `to_list` method is almost the only the reason to keep it as dataframe to the point of merging I guess...

Since we don't know where the Mtz files will be from, (I mean the software)
we should keep it flexible enough to investigate them as a data frame before we convert it to scipp object I thought.

YooSunYoung:
@SimonHeybrock  I renamed them and updated the docstring as we discussed in the chat.

src/ess/nmx/mtz_io.py
SimonHeybrock: Using `DataFrame` is perfectly fine. This comment was about adding some explanation of context in some of the `Mtz*` docstrings.

src/ess/nmx/mtz_io.py
```diff
@@ -25,14 +25,15 @@
...
 """The name of the wavelength column in the mtz file."""
 DEFAULT_WAVELENGTH_COLUMN_NAME = WavelengthColumnName("LAMBDA")
+DEFAULT_WAVELENGTH_COORD_NAME = "wavelength"
 
 IntensityColumnName = NewType("IntensityColumnName", str)
 """The name of the intensity column in the mtz file."""
 DEFAULT_INTENSITY_COLUMN_NAME = IntensityColumnName("I")
 
-SigmaIntensityColumnName = NewType("SigmaIntensityColumnName", str)
+IntensitySigColumnName = NewType("IntensitySigColumnName", str)
```
SimonHeybrock, line 34: Can we avoid abbreviations that are not very widespread, in particular if the are ambiguous? "Sig" could also mean "Signal".

How about `IntensityStdDevColumnName` or just `StdDevColumnName`? Changing `Sig` to `Sigma` would also be acceptable.

src/ess/nmx/mtz_io.py
```diff
@@ -90,15 +95,35 @@ def reduce_single_mtz(
...
+        - ``hkl``: The miller indices as a list of integers.
+        - ``d``: The d-spacing calculated from the miller indices.
+                 :math:``\\dfrac{2}{d^{2}} = \\dfrac{\\sin^2(\\theta)}{\\lambda^2}``
+        - ``resolution``: The resolution calculated from the d-spacing.
+
+        For consistent names of columns/coordinates, the following columns are renamed:
+
+        - ``wavelength_column_name`` -> ``'wavelength'``
+        - ``intensity_column_name`` -> ``'I'``
+        - ``sigma_intensity_column_name`` -> ``'SIGI'``
```
SimonHeybrock, line 123: Outdated, since renamed above.

src/ess/nmx/mtz_io.py
```diff
@@ -33,6 +48,7 @@
 """The merged mtz dataframe with derived columns."""
```
SimonHeybrock, line 48: Above: What is "Rapio"?

src/ess/nmx/mtz_io.py
```diff
@@ -193,7 +218,7 @@ def merge_mtz_dataframes(
     return MergedMtzDataFrame(pd.concat(mtz_dfs.values(), ignore_index=True))
 
 
-def reduce_merged_mtz_dataframe(
+def process_merged_mtz_dataframe(
```
SimonHeybrock, line 221: Docstring still talks about "reducing"

src/ess/nmx/mtz_io.py
```diff
@@ -227,49 +297,82 @@ def nmx_mtz_dataframe_to_scipp_dataarray(
...
+    # TODO: We can put them back once we support tuple[int] dtype.
+    # See https://github.com/scipp/scipp/issues/3046 for more details.
+    # Temporarily, we will join them into a single string.
+    # It is done on the scipp variable instead of the dataframe
+    # since columns with string dtype are converted to PyObject dtype
+    # instead of string by `from_pandas_dataframe`.
+    nmx_mtz_ds = nmx_mtz_ds.drop_coords(["hkl", "hkl_asu"])
+    nmx_mtz_ds.coords["hkl_asu"] = _join_variables(
+        *(nmx_mtz_ds.coords[f"{idx_desc}_ASU"] for idx_desc in "HKL")
+    )
```
SimonHeybrock, line 348: Why do we need do join the 3 columns previously converted form `hkl_asu` using `to_list` above? Can't w use `hkl_asu` directly? Or does it have the wrong format?

src/ess/nmx/mtz_io.py
```diff
@@ -182,19 +231,148 @@ def reduce_merged_mtz_dataframe(
...
     def _rapio_asu_to_asu(row: pd.Series) -> list[int]:
+        """Converts miller indices(HKL) to ASU indices."""
         return rapio_asu.to_asu(row["hkl"], sg.operations())[0]
 
-    merged_df["hkl_eq"] = merged_df.apply(_rapio_asu_to_asu, axis=1)
+    merged_df["hkl_asu"] = merged_df.apply(_rapio_asu_to_asu, axis=1)
+    # Unpack the indices for later.
+    merged_df[["H_ASU", "K_ASU", "L_ASU"]] = pd.DataFrame(
+        merged_df["hkl_asu"].to_list(), index=merged_df.index
+    )
```
SimonHeybrock, line 241: Correct me if I am wrong, but from very brief research I guess that this might be mapping the Miller indices into the asymmetrical unit? Isn't it still Miller indices afterwards? In particular, does it make sense to have the `asu` suffix everywhere?

src/ess/nmx/scaling.py
```diff
@@ -43,56 +36,122 @@ def get_lambda_binned(
...
+        which is selected by one ``wavelength``.
+
+    """
     # Workaround for https://github.com/scipp/scipp/issues/3046
-    grouped = ref_bin.group("H_EQ", "K_EQ", "L_EQ").flatten(
-        dims=["H_EQ", "K_EQ", "L_EQ"], to="HKL_EQ"
+    # and https://github.com/scipp/scipp/issues/3425
+    # This workaround is implemented with an assumption that
+    # the size of all combinations of (H_ASU, K_ASU, L_ASU) is small enough
+    # to be handled in memory.
```
SimonHeybrock, line 142: Couldn't we add this to the docstring of every function? Either remove it, or be more specific regarding what the particular concern here is, or whether this has been a problem in practice.

src/ess/nmx/mtz_io.py
YooSunYoung: It was because of the length limit of 88... I'll update it to `StdDevColumnName`.

src/ess/nmx/mtz_io.py
YooSunYoung: Reciprocal ASU... sorry I'll update it.

src/ess/nmx/mtz_io.py
YooSunYoung: Values of `hkl_asu` are lists.
I guess I can just convert them to string directly...

src/ess/nmx/mtz_io.py
YooSunYoung: It's just because we wanted to keep the original indices from the file, so they needed different suffix.

src/ess/nmx/mtz_io.py
YooSunYoung: When I tried converting `hkl` directly into string,
it throws this error:

```python
TypeError                                 Traceback (most recent call last)
Cell In[11], [line 1](vscode-notebook-cell:?execution_count=11&line=1)
----> [1](vscode-notebook-cell:?execution_count=11&line=1) [str(val) for val in da.coords['hkl'].values]

Cell In[11], [line 1](vscode-notebook-cell:?execution_count=11&line=1)
----> [1](vscode-notebook-cell:?execution_count=11&line=1) [str(val) for val in da.coords['hkl'].values]

TypeError: Unable to convert function return value to a Python type! The signature was
	(self: pybind11_builtins.iterator) -> scipp::python::PyObject
```

YooSunYoung: I just converted the data frame column directly though...!

src/ess/nmx/scaling.py
YooSunYoung: Actually this one should've been updated to use `hkl_asu`...

SimonHeybrock (approved):

PR #56, opened 2024-04-16
Run copier update and get the copier cron job in

MridulS:


jl-wynen:
Since we are rolling out ruff now, I think we should wait with setting up the automatic updates. This would lead to lots of conflicts.

PR #58, opened 2024-04-17
Scale intensity per HKL group per wavelength bin.

YooSunYoung:
Fixes #48 

TODO:
- [ ] Add tests
- [ ] Add plotting example in the documentation.

PR #59, opened 2024-04-25
Average scaled intensities scaled by estimated scale factors.

YooSunYoung:
Continued from #58 

I'm opening a new one since the base branch was updated a lot and there were too many conflicts to resolve...

The result of the scaling workflow now is,
the data points to fit the real scaling function with.


src/ess/nmx/scaling.py
```diff
@@ -9,23 +9,42 @@
...
+"""The number of standard deviations to be cut from the n-th root data."""
 
 # Computed types
+"""Filtered mtz dataframe by the quad root of the sample standard deviation."""
 WavelengthBinned = NewType("WavelengthBinned", sc.DataArray)
 """Binned mtz dataframe by wavelength(LAMBDA) with derived columns."""
+FilteredWavelengthBinned = NewType("FilteredWavelengthBinned", sc.DataArray)
+"""Filtered binned data."""
+SelectedReferenceWavelength = NewType("SelectedReferenceWavelength", sc.Variable)
+"""The wavelength to select reference intensities."""
```
YooSunYoung, line 32: This one had to be exposed as an intermediate result cuz it's used in later steps.

YooSunYoung:
There was one more request that allows swapping filtering function easily, (cutting out tails of the data points mostly) to compare different filtering functions.
I'll reopen it after fixing that.

YooSunYoung:
@Justin-Bergmann  and @aaronfinke briefly have reviewed in person with me.
It'll be nice if you can review them here as well.
Especially the documentation part.

You can download the built-documentation page here: https://github.com/scipp/essnmx/actions/runs/9075813196?pr=59 
There is a download button next to `docs_html` under `Artifacts` section.

src/ess/nmx/scaling.py
```diff
@@ -9,43 +9,90 @@
...
     Notes
     -----
         Wavelength(LAMBDA) binning should always be done on the merged dataset.
 
     """
-
-    return WavelengthBinned(
-        mtz_da.bin({DEFAULT_WAVELENGTH_COORD_NAME: wavelength_bin_size})
+    wavelength_coord = mtz_da.coords[DEFAULT_WAVELENGTH_COORD_NAME]
+    start = _if_not_none_else(min_wavelength_bin_edge, wavelength_coord.min())
```
jokasimr, line 77: (minor) in my opinion it's better to use the `x if x is not None else default` pattern directly here

src/ess/nmx/scaling.py
```diff
@@ -146,10 +220,144 @@ def estimate_scale_factor_per_hkl_asu_from_reference(
...
+
+    Notes
+    -----
+    The average of roughly scaled intensities are calculated by the following formula:
+
+    .. math::
+
+        EstimatedScaledI_{\\lambda}
+        = \\dfrac{
+            \\sum_{k=1}^{N_{\\lambda, (hkl)}} EstimatedScaledI_{\\lambda, (hkl)}
```
jokasimr, line 250: Does this overload `k` with two different meanings?
One as the iteration index in the sum, and one as part of the intensity peak multi-index?
Or are they actually the same `k`? In that case, why is the sum only over `k` and not over `h` and `l`? (I don't know anything about single crystal diffraction...)

src/ess/nmx/scaling.py
```diff
@@ -146,10 +220,144 @@ def estimate_scale_factor_per_hkl_asu_from_reference(
...
+            \\sum_{i=1}^{N_{reference, (hkl)}} \\dfrac{1}{I_{i}}
+        }{
+            N_{reference, (hkl)}
+        } * \\dfrac{
+            \\sum_{j=1}^{N_{\\lambda, (hkl)}} I_{j}
+        }{
+            N_{\\lambda, (hkl)}
+        } \\\\
+        = average( \\dfrac{1}{I_{ref, (hkl)}} ) * average( I_{\\lambda, (hkl)} )
+        \\end{eqnarray}
```
jokasimr, line 280: How do the arguments correspond to the quantities in the math?

What is `binned` and what is `scale_factor`?

src/ess/nmx/scaling.py
```diff
@@ -146,10 +220,144 @@ def estimate_scale_factor_per_hkl_asu_from_reference(
...
+    # to represent the average wavelength of the bin
+    # It is because the bin-edges are dropped while flattening the data
+    # and the data is expected to be filtered after this step.
+    intensities.coords[DEFAULT_WAVELENGTH_COORD_NAME] = sc.midpoints(
+        intensities.coords[DEFAULT_WAVELENGTH_COORD_NAME],
+    )
+    return EstimatedScaledIntensities(intensities)
+
+
+ScaledIntensityLeftTailThreshold = NewType(
```
jokasimr, line 304: Are these domain quantities supposed to be set by users? In that case it would be good to give them doc-strings so the user knows how to set them.

src/ess/nmx/scaling.py
```diff
@@ -146,10 +220,144 @@ def estimate_scale_factor_per_hkl_asu_from_reference(
...
+    -------
+    :
+        The filtered scaled intensities with the tails cut.
+
+    """
+    return FilteredEstimatedScaledIntensities(
+        scaled_intensities[
+            (scaled_intensities.data > left_threashold)
+            & (scaled_intensities.data < right_threshold)
+        ].copy(deep=False)
```
jokasimr, line 342: Why do we copy here?

jokasimr:
We should have a chat where you explain how this works, that would make it easier for me to review :)

src/ess/nmx/scaling.py
YooSunYoung: Yeah, it's just because of the formatting... I had make a choice among...
- make the variable names shorter and use ternary operator
- keep the long variable name even though the ternary operator will make the format ugly and unnecessarily long
- keep the long variable names make a wrapper function for the ternary operator

I valued explanatory variable names more over one less helper function.

src/ess/nmx/scaling.py
YooSunYoung: Because slicing will give you just a view, and you can't modify the view.
Copy gives you a new instance so you can work on it. F.x. adding/removing coordinate.

src/ess/nmx/scaling.py
jokasimr: I prefer option 2, keeping the names and making it longer, so something like:

```python
wavelength_coord = (
    min_wavelength_bin_edge
    if min_wavelength_bin_edge is not None else
    wavelength_coord.min()
)
```

But it is up to you to decide.

src/ess/nmx/scaling.py
```diff
@@ -146,10 +220,156 @@ def estimate_scale_factor_per_hkl_asu_from_reference(
     return EstimatedScaleFactor((1 / grouped).bins.mean())
 
 
+def average_roughly_scaled_intensities(
```
jokasimr, line 223: why is it called roughly scaled?

YooSunYoung: Because the scale factor is not the fitted one, but the estimated one.

src/ess/nmx/scaling.py
YooSunYoung: I just tried using ternary directly.
It was tedious when I was modifying the code frequently, but since it's now settled,
readability is actually not that bad in this form.
So I deleted the helper after all. Thanks : D

PR #63, opened 2024-05-27
Add fitting and scaling step in the workflow.

### LINKED ISSUE #48: [Requirement] Scaling/normalization/correction workflow
### Executive summary

We need to derive and apply scaling functions per wavelength and per position

### Context and background knowledge

There is ``scaling`` step in the whole workflow.
And it could be done by [``LSCALE``](https://onlinelibrary.wiley.com/doi/abs/10.1107/S0021889898015350),
But LSCALE is not maintained anymore so we need to implement this part ourselves.

It is for the scaling and normalization of Laue intensity data to yield fully corrected structure amplitudes. [^1]

[^1]: [``LSCALE``](https://onlinelibrary.wiley.com/doi/abs/10.1107/S0021889898015350)

### Inputs

MTZ files that contain 
- HKL
- LAMBDA
- Intensity
- Sigma of Intensity

### Methodology

This functionality was under development in [``pyscale``](https://github.com/mlund/pyscale) and there is a notebook/script in the repository.

### Outputs

MTZ file containing ...

### Which interfaces are required?

Python module / function, Python script, Jupyter notebook

### Test cases

Also available in [``pyscale``](https://github.com/mlund/pyscale) with sample data.

### Comments

 [``pyscale``](https://github.com/mlund/pyscale) is not open so you'll need to ask for an access to https://github.com/mlund/

YooSunYoung:
Fixes #48 
From the data point, we should allow users to easily swap the fitting-target-function and compare the results.

So I added a fitting step to accept a callable function as an argument and the wrapped the fitting result into a dataclass.


src/ess/nmx/scaling.py
```diff
@@ -358,6 +361,93 @@ def cut_tails(
...
+def chepyshev_polynomial(
+    wavelength: FittingInputX,
+    param1,
+    param2,
+    param3,
+    param4,
+    param5,
+    param6,
+    param7,
+) -> FittingOutputY:
```
jokasimr, line 378: Why is this function defined with so many arguments?
We're using `scipy.optimize` below anyway so there's no need to accommodate the interface of `scipp.curve_fit`

src/ess/nmx/scaling.py
```diff
@@ -358,6 +361,93 @@ def cut_tails(
...
+    initial_guess: InitialGuess,
+    fitting_func: FittingFunc = chepyshev_polynomial,
+) -> FittingResult:
+    from scipy.optimize import curve_fit
+
+    params, cov = curve_fit(
+        f=fitting_func,
+        xdata=estimated_intensities.coords["wavelength"].values,
+        ydata=estimated_intensities.data.values,
+        p0=initial_guess,
```
jokasimr, line 414: Do you have standard deviations on the ydata? Those should probably be included here through the `sigma` argument

src/ess/nmx/scaling.py
```diff
@@ -358,6 +361,93 @@ def cut_tails(
...
+    )
+    return FittingResult(
+        fitting_func=fitting_func,
+        params=FittingParams(params),
+        covariance=FittingParamsCovariances(cov),
+        fit_output=sc.DataArray(
+            data=sc.array(
+                dims=["wavelength"],
+                values=fitting_func(
+                    estimated_intensities.coords["wavelength"].values, *params
```
jokasimr, line 424: If you add variances to the `params` then the `fit_output` will also have uncertainties
This is done automatically in `scipp.curve_fit`.

jokasimr (commented):
I don't really understand why we are using such high polynomial degree here. And why Chebychev polynomials instead of something like `ax**3 + bx**2 + c*x + d` or `np.polyval((a, b, c, d), x)`?
The advantage of `ax**3 + bx**2 + c*x + d` is that the fitting function can be implemented in scipp.
From the scatter plots I've seen before they seem noisy so that fitting anything above degree 3 is likely just noise.
But maybe those were not typical.
I think it's also worth considering other ["loss functions"](https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.least_squares.html#scipy.optimize.least_squares) if there are outliers in the data.

src/ess/nmx/scaling.py
```diff
@@ -358,6 +361,93 @@ def cut_tails(
     )
 
 
+FittingInputX = NewType("FittingInputX", sc.Variable)
+FittingOutputY = NewType("FittingOutputY", Sequence)
+FittingFunc = Callable[..., FittingOutputY]
+
+
+def chepyshev_polynomial(
```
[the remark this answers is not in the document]
YooSunYoung: It's just a typo not intentional... : D...

src/ess/nmx/scaling.py
YooSunYoung: Yeah... I was hoping that we can use `scipp.curve_fit` when writing this... I'll update it.

src/ess/nmx/scaling.py
YooSunYoung: Hmm... it diverges to infinite with a sample data....

YooSunYoung:
> I don't really understand why we are using such high polynomial degree here. And why Chebychev polynomials instead of something like `ax**3 + bx**2 + c*x + d` or `np.polyval((a, b, c, d), x)`? The advantage of `ax**3 + bx**2 + c*x + d` is that the fitting function can be implemented in scipp.
> From the scatter plots I've seen before they seem noisy so that fitting anything above degree 3 is likely just noise. But maybe those were not typical?
> 

I don't know why it was being used but this ``fitting_func`` is meant to be swapped
so I just picked the most complicated one, to make sure that it works with all different fitting functions that are used in the original scaling routine.

Also don't know how the real data will look like... 

I think we can't decide actual fitting functions and optimization methods here right now,
so I was trying to make it generic as possible.

> I think it's also worth considering other ["loss functions"](https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.least_squares.html#scipy.optimize.least_squares) if there are outliers in the data.

Good suggestion...! 
At this point, we were hoping that most of outliers are filtered, but the data doesn't seem so.

src/ess/nmx/scaling.py
YooSunYoung: When I try to directly use the `fitting_func`, in this case `Chebyshev`, 
it causes unit error... and I'm not sure how to fix this... 

```
File ~/ESS/essnmx/src/ess/nmx/scaling.py:370, in chebyshev_polynomial(wavelength, *params)
    [369](https://file+.vscode-resource.vscode-cdn.net/Users/sunyoungyoo/ESS/essnmx/docs/examples/~/ESS/essnmx/src/ess/nmx/scaling.py:369) def chebyshev_polynomial(wavelength: FittingInputX, *params) -> FittingOutputY:
--> [370](https://file+.vscode-resource.vscode-cdn.net/Users/sunyoungyoo/ESS/essnmx/docs/examples/~/ESS/essnmx/src/ess/nmx/scaling.py:370)     return np.polynomial.chebyshev.Chebyshev(params)(wavelength)

File ~/miniconda3/envs/nmx-dev-310/lib/python3.10/site-packages/numpy/polynomial/_polybase.py:512, in ABCPolyBase.__call__(self, arg)
    [510](https://file+.vscode-resource.vscode-cdn.net/Users/sunyoungyoo/ESS/essnmx/docs/examples/~/miniconda3/envs/nmx-dev-310/lib/python3.10/site-packages/numpy/polynomial/_polybase.py:510) def __call__(self, arg):
    [511](https://file+.vscode-resource.vscode-cdn.net/Users/sunyoungyoo/ESS/essnmx/docs/examples/~/miniconda3/envs/nmx-dev-310/lib/python3.10/site-packages/numpy/polynomial/_polybase.py:511)     off, scl = pu.mapparms(self.domain, self.window)
--> [512](https://file+.vscode-resource.vscode-cdn.net/Users/sunyoungyoo/ESS/essnmx/docs/examples/~/miniconda3/envs/nmx-dev-310/lib/python3.10/site-packages/numpy/polynomial/_polybase.py:512)     arg = off + scl*arg
    [513](https://file+.vscode-resource.vscode-cdn.net/Users/sunyoungyoo/ESS/essnmx/docs/examples/~/miniconda3/envs/nmx-dev-310/lib/python3.10/site-packages/numpy/polynomial/_polybase.py:513)     return self._val(arg, self.coef)
```

src/ess/nmx/scaling.py
```diff
@@ -362,30 +361,36 @@ def cut_tails(
...
+    param4: sc.Variable,
+    param5: sc.Variable,
+    param6: sc.Variable,
+    param7: sc.Variable,
 ) -> FittingOutputY:
-    return np.polynomial.chebyshev.Chebyshev(
-        (param1, param2, param3, param4, param5, param6, param7)
-    )(wavelength)
+    return FittingOutputY(
+        (wavelength**6) * sc.values(param7)
```
jokasimr, line 379: Why `sc.values` here? Might be a good idea to multiply by `sc.units.angstrom**(-k)` to make the parameters dimensionless, just for convenience, otherwise you have pass initial guesses for them in the call to `curve_fit`.

src/ess/nmx/scaling.py
```diff
@@ -362,30 +361,36 @@ def cut_tails(
...
+        + sc.values(param1)
+    )
 
 
 FittingParams = NewType("FittingParams", Mapping)
 """Parameters of the fitting function."""
 FittingParamsCovariances = NewType("FittingParamsCovariances", Mapping)
 """Covariance of the :attr:`~FittingParams`."""
-InitialGuess = NewType("InitialGuess", np.ndarray)
+InitialGuess = NewType("InitialGuess", dict)
```
jokasimr, line 393: I think we shouldn't have an initial guess input. Unless you've encountered cases when the default initial guess was insufficient.

src/ess/nmx/scaling.py
YooSunYoung: I wasn't sure if it's a good idea to wrap the parameters like that or not.

src/ess/nmx/scaling.py
YooSunYoung: No, just debating between wrap the parameters to have a write unit in the fitting function or not...

src/ess/nmx/scaling.py
YooSunYoung: I'll... decide it tomorrow...! : D

src/ess/nmx/scaling.py
```diff
@@ -358,6 +360,91 @@ def cut_tails(
...
+        da=estimated_intensities,
+        p0=initial_guess,
+    )
+    data = fitting_func(estimated_intensities.coords["wavelength"], **p_result)
+    return FittingResult(
+        fitting_func=fitting_func,
+        params=FittingParams(p_result),
+        covariance=FittingParamsCovariances(cov_result),
+        fit_output=sc.DataArray(
+            data=data.data if isinstance(data, sc.DataArray) else data,
```
jokasimr, line 425: Is `data` not always a `DataArray`? I think it should be, since the parameters in `p_result` are `DataArrays`

src/ess/nmx/scaling.py
jokasimr: Since we are using `scipp.curve_fit` now it makes sense to have it like this.

If we **use the latest version of scipp** there is a method we can use to work around some of the inconvenience of not being able to pass a variable number of positional arguments to `scipp.curve_fit` we could do something like this:

```python
def polyval(x, **kwargs):
    out = 0.0
    xk = sc.ones_like(x)
    for a in kwargs.values():
        out += a * xk
        xk *= x
    return out
```
and then in the `curve_fit` we can pass a variable number of parameters like this:
```python
par, _ = sc.curve_fit(
    ['x'],
    polyval,
    da,
    p0={f'a{i}': 1.0 for i in range(degree_of_polynomial_user_requested+1)},
)
```
(instead of `1.0` here you would maybe need to use a guess that makes sense unit-wise).

`par`  will be a dataarray that looks like:
```
DataGroup(sizes={}, keys=[
    a0: DataArray({}),
    a1: DataArray({}),
    a2: DataArray({}),
])
```
and we can evaluate the fit function at the parameter values like `polyval(x, **par)`.

src/ess/nmx/scaling.py
```diff
@@ -358,6 +360,123 @@ def cut_tails(
...
+    :
+        The polynomial calculated at the wavelength.
+
+
+    """
+    out = sc.zeros_like(wavelength)
+    out.unit = out_unit
+    xk = sc.ones_like(wavelength)
+    for _, arg_value in enumerate(kwargs.values()):
+        out += sc.values(arg_value) * xk * sc.scalar(1.0, unit=out.unit / xk.unit)
```
jokasimr, line 405: Would it be more efficient if we moved the unit inside the first call to `sc.scalar`, like this:
```python
out += sc.values(arg_value, unit=out.unit / xk.unit) * xk 
```
Then we avoid one multiplication with the array `xk`

jokasimr (approved):
Looks good! Left one comment

src/ess/nmx/scaling.py
YooSunYoung: Oh I didn't know `values` allow unit conversion. I'll do that.

YooSunYoung: Ah, no it doesn't work : D...

jokasimr: I read `scalar` instead of `values` 😅

PR #73, opened 2024-06-12
Refactor for next Sciline

### LINKED ISSUE #72: Space group description overrides info from files?
Docstrings says https://github.com/scipp/essnmx/blob/611c89a6534e1c6cbb9e65c6f7705955e761a316/src/ess/nmx/mtz_io.py#L173-L175 but https://github.com/scipp/essnmx/blob/611c89a6534e1c6cbb9e65c6f7705955e761a316/src/ess/nmx/mtz_io.py#L194-L203 seems to *override* any info from files with the provided description.

Furthermore, the clean that `DEFAULT_SPACE_GROUP_DESC` is used is incorrect.

If I naively change this to match what the docstring claims, I get `ValueError: Multiple space groups found: [<gemmi.SpaceGroup("P 21 21 21")>, <gemmi.SpaceGroup("P 21 21 21")>, <gemmi.SpaceGroup("P 21 21 21")>, <gemmi.SpaceGroup("P 21 21 21")>, <gemmi.SpaceGroup("P 21 21 21")>]` when running the docs workflow, I suppose since all files have the same space group.

SimonHeybrock:
Preparation for next Sciline release. Note that this currently depends on `sciline@main`, which needs updating when Sciline is released. The plan is to make a synchronized release.

Fixes #72. Is this fix correct?


docs/examples/workflow.ipynb
```diff
@@ -54,6 +51,28 @@
...
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# DetectorIndex selects what detector panels to include in the run\n",
+    "# in this case we select all three panels.\n",
+    "wf[NMXReducedData] = (\n",
+    "    wf[NMXReducedData]\n",
+    "    .map(pd.DataFrame({DetectorIndex: range(3)}).rename_axis(\"panel\"))\n",
```
jokasimr, line 71: How come pandas is used here now? I thought pandas was only used in this repo for the mtz stuff.

SimonHeybrock: It is one way of defining values to map over. The advantage of Pandas is that (a) users are likely to be familiar with it, and (b) can define an index name ("panel" above), which is not possible currently when using a plain `dict`.

jokasimr: I mean, users are likely to be familiar with a plain dict as well. Is using `pd.DataFrame` the only way to add an index name?

SimonHeybrock: Alternatively, we could use a dict of `scipp.Variable`.

jokasimr: Since we are already using scipp everywhere else maybe it's better to use that here as well? I think mixing in packages for no reason is confusing.

jokasimr (approved):

PR #76, opened 2024-07-04
Copier update

YooSunYoung:
We haven't updated copier for a while...! : D.... And that led to the broken version in the docs...


YooSunYoung:
#77  should be approved first.

MridulS (approved):

PR #85, opened 2024-09-27
Rotated detector pixel offset based on the starting position.

YooSunYoung:
We were assuming the first pixel is always at `(0,0,0)` but it was not the case.

I added this to be included.

**I didn't clean up the code or write more tests because this geometry parsing will be obsolete soon.**

### DIFF
--- a/src/ess/nmx/mcstas/xml.py
+++ b/src/ess/nmx/mcstas/xml.py
@@ -338,8 +338,11 @@ def _pixel_positions(
     return (
         (pixel_n_slow * slow_axis_steps)
         + (pixel_n_fast * fast_axis_steps)
-        + position_offset
-    )
+        + detector.rotation_matrix
+        * sc.vector(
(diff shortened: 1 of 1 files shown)


jokasimr (commented):
Are the `position_offset` vectors the offsets of the pixels from the origin of the detector coordinate system? In that case, should they not be rotated with the detector? 

Is the detector rotation matrix a rotation around the detectors own axis, or a rotation of the detector around the sample?

jokasimr:
@SimonHeybrock suggested merging this even if there are some question marks since it's mainly for testing anyway.

jokasimr (approved):

YooSunYoung:
> Are the `position_offset` vectors the offsets of the pixels from the origin of the detector coordinate system? In that case, should they not be rotated with the detector?


Yes. `position_offset` is not rotated.

> Is the detector rotation matrix a rotation around the detectors own axis, or a rotation of the detector around the sample?

It's a rotation around the detector's own axis.

PR #87, opened 2024-10-15
Add Python 3.13 to supported versions, copier update, tox deps bump

MridulS:


nvaytet (commented):
Changes look good. @MridulS fixing the CI should be easy enough?

PR #88, opened 2024-10-29
McStas 3.5 version workaround.

YooSunYoung:
It's just for supporting Aaron quickly for the new mcstas simulation.
The new implementation might change again so I would rather not put more effort than this for now.
So plz ignore this PR : D


YooSunYoung:
Still not clear about this. We'll follow it up later...!

PR #93, opened 2024-11-25
Update workflow user guide to use latest sciline interface

YooSunYoung:
I included some subtle steps of `map`/`reduce` with short explanation since IDS member wanted to know the details as well.
This workflow will be the first/basic workflow whenever new member joins for NMX team, so I think it's okay to be verbose a bit...?


SimonHeybrock:
I think we were going to discuss the panel mapping anyway, should we put this on hold until then?

YooSunYoung:
> I think we were going to discuss the panel mapping anyway, should we put this on hold until then?

Yes, I prepared this for that discussion so that we can start from here. I don't mind holding it until then.

YooSunYoung:
This one is obsolete now.

PR #99, opened 2025-01-24
Add explanation in the comment of the bin edge parameter.

YooSunYoung:
Fixes scipp/ess#211 

output

Also removes unnecessary stuff.


YooSunYoung:
Closing this since #100 fixes the root cause of the problem.

PR #100, opened 2025-01-24
fix: use float64 to avoid truncation

### LINKED ISSUE #97: Fitting function returns different results for different versions of scipy
# With Scipy version 1.12.0
Image

# With Scipy version 1.15.0
Image

We're still not sure what exactly caused the changes....

jokasimr:
Fixes #97

### DIFF
--- a/src/ess/nmx/scaling.py
+++ b/src/ess/nmx/scaling.py
@@ -359,6 +359,9 @@ def polyval_wavelength(
 
 
     """
+    # We need to use float64 precision because
+    # the curve fit routine depends on finite difference
+    # estimates of the derivative.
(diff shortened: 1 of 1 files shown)


src/ess/nmx/scaling.py
```diff
@@ -359,6 +359,7 @@ def polyval_wavelength(
 
 
     """
+    wavelength = wavelength.to(dtype='float64')
```
YooSunYoung, line 362: Can we add a comment here explaining why we're doing this?

YooSunYoung (approved):
I'll close the other PR since this one fixing more lower level issue.

PR #101, opened 2025-01-27
Export dataset according to NXLauetof application definition

YooSunYoung:
Related to #60 

This PR only contains place holders.
Due to memory issue, we decided to split exporting function into panel-dependent one and panel-independent one.

### DIFF
--- a/src/ess/nmx/nexus.py
+++ b/src/ess/nmx/nexus.py
@@ -129,6 +129,15 @@ def export_as_nexus(
 
     Currently exporting step is not expected to be part of sciline pipelines.
     """
+    import warnings
+
+    warnings.warn(
+        DeprecationWarning(
+            "Exporting to custom NeXus format will be deprecated in the near future."
+            "Please use ``export_as_nxlauetof`` instead."
+        ),
+        stacklevel=1,
+    )
     with h5py.File(output_file, "w") as f:
         f.attrs["default"] = "NMX_data"
         nx_entry = _create_root_data_entry(f)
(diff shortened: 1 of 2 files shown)


src/ess/nmx/nexus.py
```diff
@@ -129,10 +129,173 @@ def export_as_nexus(
...
+
+def _create_lauetof_data_entry(file_obj: h5py.File) -> h5py.Group:
+    nx_entry = file_obj.create_group("entry")
+    nx_entry.attrs["NX_class"] = "NXentry"
+    return nx_entry
+
+
+def _add_lauetof_definition(nx_entry: h5py.Group) -> None:
+    nx_entry["definition"] = "NXlauetof"
+    nx_entry["definition"].attrs["NX_class"] = "NX_CHAR"
```
SimonHeybrock, line 158: I don't think fields in NeXus should have an `NX_class` attribute?

src/ess/nmx/nexus.py
```diff
@@ -129,10 +129,173 @@ def export_as_nexus(
...
+        # Skipping ``name`` field
+
+
+def export_panel_dependent_data_as_nxlauetof(
+    *dgs: sc.DataGroup, output_file: str | pathlib.Path | io.BytesIO
+) -> None:
+    with h5py.File(output_file, "r+") as f:
+        nx_instrument: h5py.Group = f["entry/instrument"]
+        for dg in dgs:
+            _add_lauetof_detector_group(dg, nx_instrument)
```
SimonHeybrock, line 291: Not sure I understand. If `dgs` contains all panels, we still use 3x the memory (leaving aside intermediate steps). Wasn't the point to have a writer that works one panel at a time?

src/ess/nmx/nexus.py
YooSunYoung: It was specified in the application definition but I wasn't sure if it's the right way. Do you think we should remove them?

src/ess/nmx/nexus.py
YooSunYoung: I thought you might also want to write them all at the same time, if they are not too large.
Fir example, you need all three panels at the same time if you want to see all three of them in an instrument view anyways.

src/ess/nmx/nexus.py
SimonHeybrock: Can't see it here: https://manual.nexusformat.org/classes/applications/NXlauetof.html. Can you give a reference?

YooSunYoung: Ah it was DTYPE not NX_class sorry. I was confused cause they were written next to the field... I removed it.

src/ess/nmx/nexus.py
SimonHeybrock: "I thought you might also want" is often an indicator that it should be postponed until we know it is really needed. Otherwise you need to refactor three times and write tests. And I am not sure why the instrument view is related to this.

In any case: How do you imagine the interface for writing multiple panels to the same file when they are *not* in memory at the same time?

YooSunYoung: > And I am not sure why the instrument view is related to this.

After using instrument view, there's no point of iterating it again and you can just save them all at once.

> In any case: How do you imagine the interface for writing multiple panels to the same file when they are not in memory at the same time?

The idea was to save panel independent data first always and then append panel-dependent data into the same file.
```python

file_name = "test.nxs"
export_panel_independent_data_as_nxlauetof(binned_dg, output_file=file_name)
export_panel_dependent_data_as_nxlauetof(binned_dg, output_file=file_name)
del binned_dg

for i in range(1, 3):
    with temp_parameter(wf, DetectorIndex, i) as temp_wf:
        reduced_data = temp_wf.compute(NMXReducedData)
        export_panel_dependent_data_as_nxlauetof(reduced_data, output_file=file_name)
        del reduced_data
```

And if you want to save them all at the same time, you can just
```python
```

SimonHeybrock: Does the panel-independent data need the workflow? If so, how do we handle that?

YooSunYoung: > panel-independent data need the workflow

Monitor(control) group might need its own workflow, (we haven't seen any use case of monitor data yet)
but other than that, all other fields only needs a loader.

But also same information is available from the reduced data.

PR #109, opened 2025-02-25
Update workflow to be able to process chunk by chunk.

YooSunYoung:
Some of steps needed to be split/reorganized to process big file chunk by chunk.

splitting/cleaning up the two other PRs https://github.com/scipp/essnmx/pull/102 and https://github.com/scipp/essnmx/pull/101

### DIFF
--- a/docs/examples/workflow.ipynb
+++ b/docs/examples/workflow.ipynb
@@ -23,7 +23,7 @@
     "from ess.nmx.data import small_mcstas_3_sample\n",
     "\n",
     "from ess.nmx.types import *\n",
-    "from ess.nmx.reduction import NMXData, NMXReducedData, merge_panels\n",
+    "from ess.nmx.reduction import merge_panels\n",
     "from ess.nmx.nexus import export_as_nexus\n",
     "\n",
     "wf = McStasWorkflow()\n",
@@ -64,8 +64,8 @@
    "source": [
     "# DetectorIndex selects what detector panels to include in the run\n",
     "# in this case we select all three panels.\n",
-    "wf[NMXReducedData] = (\n",
-    "    wf[NMXReducedData]\n",
+    "wf[NMXReducedDataGroup] = (\n",
+    "    wf[NMXReducedDataGroup]\n",
     "    .map({DetectorIndex: sc.arange('panel', 3, unit=None)})\n",
     "    .reduce(index=\"panel\", func=merge_panels)\n",
     ")"
@@ -84,7 +84,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "wf.visualize(NMXReducedData, graph_attr={\"rankdir\": \"TD\"}, compact=True)"
+    "wf.visualize(NMXReducedDataGroup, graph_attr={\"rankdir\": \"TD\"}, compact=True)"
    ]
   },
   {
@@ -102,8 +102,8 @@
    "source": [
     "from cyclebane.graph import NodeName, IndexValues\n",
     "\n",
-    "# Event data grouped by pixel id for each of the selected detectors\n",
-    "targets = [NodeName(NMXData, IndexValues((\"panel\",), (i,))) for i in range(3)]\n",
+    "# Data from all selected detectors binned by panel, pixel and timeslice\n",
+    "targets = [NodeName(NMXReducedDataGroup, IndexValues((\"panel\",), (i,))) for i in range(3)]\n",
     "dg = merge_panels(*wf.compute(targets).values())\n",
     "dg"
    ]
@@ -114,9 +114,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "# Data from all selected detectors binned by panel, pixel and timeslice\n",
-    "binned_dg = wf.compute(NMXReducedData)\n",
-    "binned_dg"
+    "dg['counts']"
    ]
   },
   {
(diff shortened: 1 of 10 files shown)


src/ess/nmx/mcstas/load.py
```diff
@@ -34,40 +34,113 @@ def load_event_data_bank_name(
...
+        root = f["entry1/data"]
+        dset = root[bank_name]["events"]
+        if chunk_size == 0:
+            for data_slice in dset.dataset.iter_chunks():
+                yield _wrap_raw_event_data(dset[data_slice])
+        else:
+            num_events = dset.shape[0]
+            for start in range(0, num_events, chunk_size):
+                data = dset["dim_0", start : start + chunk_size]
+                yield _wrap_raw_event_data(data)
```
YooSunYoung, line 143: Only this one is the new interface to load data chunk by chunk.
Rest is about reorganizing workflow.

docs/examples/workflow.ipynb
```diff
@@ -64,8 +64,8 @@
    "source": [
     "# DetectorIndex selects what detector panels to include in the run\n",
     "# in this case we select all three panels.\n",
-    "wf[NMXReducedData] = (\n",
-    "    wf[NMXReducedData]\n",
+    "wf[NMXReducedDataGroup] = (\n",
```
jokasimr, line 67: What's the difference between `NMXReducedData` and `NMXReducedDataGroup`?

docs/examples/workflow.ipynb
```diff
@@ -102,8 +102,8 @@
    "source": [
     "from cyclebane.graph import NodeName, IndexValues\n",
     "\n",
-    "# Event data grouped by pixel id for each of the selected detectors\n",
-    "targets = [NodeName(NMXData, IndexValues((\"panel\",), (i,))) for i in range(3)]\n",
+    "# Data from all selected detectors binned by panel, pixel and timeslice\n",
```
jokasimr, line 105: Sounds like that might be a lot of bins, how many bins are there typically?
Is the time dimension a binning of the time of flight or is it a binning of the experiment time?

src/ess/nmx/mcstas/load.py
```diff
@@ -34,40 +34,113 @@ def load_event_data_bank_name(
...
+    )
+
+
+def _exclude_zero_events(data: sc.Variable) -> sc.Variable:
+    """Exclude events with zero counts from the data.
+
+    McStas can add an extra event line containing 0,0,0,0,0,0
+    This line should not be included so we skip it.
+    """
+    if (data.values[0] == 0).all():
```
jokasimr, line 50: I remember there was some talk that not only the first row in the data might be all zeros, is that the case?
If so we probably need to check the other rows as well, not just the first.

src/ess/nmx/mcstas/load.py
jokasimr: Can we apply the rest of the workflow to the generated event chunks?

docs/examples/workflow.ipynb
YooSunYoung: `*ReducedData` only has the data array (event data or histogram) and `*ReducedDataGroup` has all other metadata,
i.e. proton charge, detector-name etc.

docs/examples/workflow.ipynb
YooSunYoung: 3 * (1280*1280) * (50~100) 
so... ~10 million?
It's binning of the time of flight.
DIALS need time of flight, not the wavelength so we use `t` for `time-of-flight`

src/ess/nmx/mcstas/load.py
YooSunYoung: I thought that issue was fixed, no...?

docs/examples/workflow.ipynb
jokasimr: `3 * (1280*1280) * 50 = 245_760_000` = 5 GB to store the bin start and end indices

Seems manageable :+1:

jokasimr: When reducing chunks it's probably good to read chunks that are at least this big.
Because otherwise there will be very few events compared to the number of bins.

YooSunYoung: I just checked and 10_000_000 is faster than 100_000_000 
so I added a warning if the chunk size is less than 10_000_000

YooSunYoung: I accidentally pushed it into PR #111 ....

jokasimr: > I just checked and 10_000_000 is faster than 100_000_000

Good test! I wonder why that's the case.

jokasimr: What was the time difference?

jokasimr (approved):

docs/examples/workflow.ipynb
YooSunYoung: Just tried once more each.
It probably also depends on the machine.

# Computing Maximum Probability (No binning)
| Chunk Size | Time [s] | 
| - | - |
| 10_000_000 | 8.5 | 
| 100_000_000 | 19.4 |

# Computing histogram (Binning into pixel id and time of flight)
| Chunk Size | Time [s] | 
| - | - |
| 10_000_000 | 22.1 | 
| 100_000_000 | 32.5 |

YooSunYoung: Ah, but the total events were just 139_494_365. Hmm...

PR #110, opened 2025-02-26
Add essreduce as a dependency to use stream processor

YooSunYoung:


pyproject.toml
```diff
@@ -32,6 +32,7 @@ requires-python = ">=3.10"
 # Make sure to list one dependency per line.
 dependencies = [
   "dask",
+  "essreduce>=24.02.3",
```
YooSunYoung, line 35: This is the only manually added dependency.

MridulS (approved):

PR #111, opened 2025-02-26
Chunk data processing example.

YooSunYoung:


src/ess/nmx/streaming.py
```diff
@@ -0,0 +1,70 @@
...
+    @property
+    def value(self) -> sc.Variable | None:
+        return self._cur_max
+
+    def _do_push(self, value: sc.Variable) -> None:
+        new_max = value.max()
+        if self._cur_max is None:
+            self._cur_max = new_max
+        else:
+            self._cur_max = sc.concat([self._cur_max, new_max], dim='max').max('max')
```
jokasimr, line 30: (minor) it's more concise to use `min()` here if possible

jokasimr (approved):
While this works, and it seems to run faster than the other script, I agree with what you said earlier @YooSunYoung that some of what we are doing here is maybe overkill for what Aaron wants done. What do you think?

YooSunYoung:
> While this works, and it seems to run considerably faster than the other script, I agree with what you said earlier @YooSunYoung that some of what we are doing here is maybe overkill for what Aaron wants done. What do you think?

I do think it's overkill but also a good example.

So I'd say it doesn't matter if we decide to write a script in a flat/plain way, not using stream processor and sciline, we just keep this example here as a documentation.

It's easier to explain what it's supposed to do in this way.
And it'll be useful if someone wants to use notebook after all.

PR #114, opened 2025-02-27
Retrieving raw data metadata separately

YooSunYoung:
We would like to retrieve some statistics, i.e. min/max of time of arrival from the data without reducing the data.

So I separated this step out of the loader.

### DIFF
--- a/src/ess/nmx/mcstas/load.py
+++ b/src/ess/nmx/mcstas/load.py
@@ -288,16 +288,6 @@ def retrieve_pixel_ids(
     return PixelIds(instrument.pixel_ids(detector_name))
 
 
-def calculate_minimum_toa(da: RawEventProbability) -> MinimumTimeOfArrival:
-    """Calculate the minimum time of arrival from the data."""
-    return MinimumTimeOfArrival(da.coords['t'].min())
-
-
-def calculate_maximum_toa(da: RawEventProbability) -> MaximumTimeOfArrival:
-    """Calculate the maximum time of arrival from the data."""
-    return MaximumTimeOfArrival(da.coords['t'].max())
-
-
 def retrieve_raw_data_metadata(
     min_toa: MinimumTimeOfArrival,
     max_toa: MaximumTimeOfArrival,
@@ -314,8 +304,6 @@ def retrieve_raw_data_metadata(
 
 
 providers = (
-    calculate_minimum_toa,
-    calculate_maximum_toa,
     retrieve_raw_data_metadata,
     read_mcstas_geometry_xml,
     detector_name_from_index,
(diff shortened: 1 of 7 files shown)


src/ess/nmx/mcstas/load.py
```diff
@@ -245,22 +250,34 @@ def bank_names_to_detector_names(description: str) -> dict[str, list[str]]:
...
+    )
+
+
+def load_detector_metadata(
+    instrument: McStasInstrument, detector_name: DetectorName
+) -> NMXDetectorMetadata:
+    """Load the detector metadata from the McStas file."""
+    return NMXDetectorMetadata(
+        sc.DataGroup(**instrument.detector_metadata(detector_name))
+    )
```
jl-wynen, line 270: These functions don't look like they load anything. They just assemble already loaded data. Is there a way to load only the requested (meta)data? This is how the generic nexus workflow works.

src/ess/nmx/mcstas/load.py
```diff
@@ -271,7 +288,35 @@ def retrieve_pixel_ids(
...
 
 
+def calculate_minimum_toa(da: RawEventProbability) -> MinimumTimeOfArrival:
+    """Calculate the minimum time of arrival from the data."""
+    return MinimumTimeOfArrival(da.coords['t'].min())
+
+
+def calculate_maximum_toa(da: RawEventProbability) -> MaximumTimeOfArrival:
+    """Calculate the maximum time of arrival from the data."""
+    return MaximumTimeOfArrival(da.coords['t'].max())
```
jl-wynen, line 298: This is not part of loading data. It is a processing step for data that is already in memory. Plus, this is no specific to McStas. So can you move this out of `load.py`?

src/ess/nmx/mcstas/load.py
YooSunYoung: ``McStasInstrument`` works as a bridge to load/calculate/wrap metadata.
The implementation ended up like this because McStas was still expecting some breaking changes.

src/ess/nmx/mcstas/load.py
YooSunYoung: It is specific to McStas because it's for calculating ``McStasWeight2CountScaleFactor`` and ``ProtonCharge`` which are not necessary for real data.

But I can move it out of the loader though.

src/ess/nmx/mcstas/load.py
YooSunYoung: At the moment most of the metadata is from one xml object so... no, we can't load just the part of the metadata.

jl-wynen (approved):

PR #115, opened 2025-02-27
Lauetof format export function

YooSunYoung:
Rewriting #102 and #101 

Fixes #96

PR #116, opened 2025-02-27
Validity check when appending histogram on top of metadata 

YooSunYoung:
For now it's only checking very few things but later we can add more metadata in the exported group and check them later.


jokasimr (commented):
Why do we need to check that the metadata matches? Is it plausible that they don't match?

YooSunYoung:
> Why do we need to check that the metadata matches? Is it plausible that they don't match?

Yes, you could in principle use any files to append the histogram if detector name matches.

jokasimr (approved):

SimonHeybrock:
Related: https://github.com/scipp/essnmx/pull/115#discussion_r1983111334

PR #118, opened 2025-02-27
Data reduction wrapper interface

YooSunYoung:
This interface was requested by IDS member since it is expected to be run on a cluster, not in a notebook.


YooSunYoung:
@jokasimr we can try the comparison tomorrow...!

src/ess/nmx/mcstas/executables.py
```diff
@@ -0,0 +1,268 @@
...
+    # Set the crystal rotation manually for now ...
+    wf[CrystalRotation] = sc.vector([0, 0, 0.0], unit='deg')
+    reduction(
+        input_file=input_file,
+        output_file=output_file,
+        chunk_size=args.chunk_size,
+        detector_ids=args.detector_ids,
+        logger=logger,
+        wf=wf,
+    )
```
YooSunYoung, line 268: Hi @aaronfinke ...! Sorry for such a delay... 
This part should do the same job as the `process.py` you showed me before.

It doesn't have performance counter, but we can add performance counting if you want.
For 17GB of data, it took 40 seconds with chunk size of 10_000_000

Here is a simple benchmark results I collected so far:

# ~1e8 events per detector performance
| chunk size | time [s] |
| - | - |
| 1_000_000 | 128 | 
| 10_000_000 | 40 |
| 50_000_000 | 34 |
| 100_000_000 | 48 |

requirements/base.txt
```diff
@@ -29,7 +29,7 @@ dnspython==2.7.0
     # via email-validator
 email-validator==2.2.0
     # via scippneutron
-essreduce==25.2.4
+essreduce==25.2.5
```
YooSunYoung, line 32: The newer version of essreduce had a breaking change in the accumulator interface.

I should also add it in the nightly dependency.

src/ess/nmx/mcstas/executables.py
SimonHeybrock: Do we know what is the slow part? Is it the McStas load part? I am used to a bit better performance (at least when data is on an SSD).

YooSunYoung: I haven't done a thorough profiling.
This benchmark includes loading the data on the memory and also writing the output into a file.

aaronfinke:
What would also be useful is a way to distinguish between McStas simulated processed data and (what will be) processed real data. I suggest either putting "McStas" in the name entry `/entry/sample/name` or perhaps as an attribute in `/entry/instrument.attrs('source') = "McStas"` or both.

src/ess/nmx/mcstas/executables.py
```diff
@@ -0,0 +1,268 @@
...
+
+    if logger is not None:
+        logger.info("Exporting metadata into the output file %s", output_file)
+
+    export_metadata_as_nxlauetof(
+        *detector_metas,
+        experiment_metadata=experiment_metadata,
+        output_file=output_file,
+        # Arbitrary metadata falls into ``entry`` group as a variable.
+        mcstas_weight2count_scale_factor=scale_factor,
```
YooSunYoung, line 173: @aaronfinke
About the mcstas indicator, we also need to export mcstas specific parameters like this.
Currently if you use this particular interface, you can freely add arbitrary metadata dataset. 
But maybe attributes are more convenient and easy to read... I'll work on it: #121

YooSunYoung:
Continued in #122

PR #122, opened 2025-03-11
Data reduction wrapper interface (cleaned)

YooSunYoung:
#115  had too many changes to keep up so I just moved the changes from #118 on the clean state.

Continues from #118

### DIFF
--- a/src/ess/nmx/mcstas/executables.py
+++ b/src/ess/nmx/mcstas/executables.py
@@ -1,0 +1,277 @@
+# SPDX-License-Identifier: BSD-3-Clause
+# Copyright (c) 2025 Scipp contributors (https://github.com/scipp)
+import argparse
+import logging
+import pathlib
+import sys
+from collections.abc import Callable
+from functools import partial
+
+import sciline as sl
+import scipp as sc
+
+from ess.reduce.streaming import (
+    EternalAccumulator,
+    MaxAccumulator,
+    MinAccumulator,
+    StreamProcessor,
+)
+
+from ..nexus import (
+    _export_detector_metadata_as_nxlauetof,
+    _export_reduced_data_as_nxlauetof,
+    _export_static_metadata_as_nxlauetof,
+)
+from ..streaming import calculate_number_of_chunks
+from ..types import (
+    CrystalRotation,
+    DetectorIndex,
+    DetectorName,
+    FilePath,
+    MaximumCounts,
+    MaximumProbability,
+    MaximumTimeOfArrival,
+    McStasWeight2CountScaleFactor,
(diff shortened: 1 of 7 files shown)


src/ess/nmx/mcstas/executables.py
```diff
@@ -0,0 +1,274 @@
...
+        min_toa=min_toa, max_toa=max_toa, max_probability=max_probability
+    )
+
+
+def reduction(
+    *,
+    input_file: pathlib.Path,
+    output_file: pathlib.Path,
+    chunk_size: int = 10_000_000,
+    detector_ids: list[int],
```
MridulS, line 142:
```suggestion
    detector_ids: list[int | str],
``` 
?
The type checkers seems to be complaining about `detector_ids` too as the `main` function forces it to be integer IDs and not detector names.

src/ess/nmx/mcstas/executables.py
```diff
@@ -0,0 +1,274 @@
...
+    logger: logging.Logger | None = None,
+) -> None:
+    wf = wf.copy() if wf is not None else McStasWorkflow()
+    wf[FilePath] = input_file
+    # Set static info
+    wf[McStasInstrument] = wf.compute(McStasInstrument)
+
+    # Calculate parameters for data reduction
+    data_metadata = calculate_raw_data_metadata(
+        *detector_ids, wf=wf, logger=logger, chunk_size=chunk_size
```
MridulS, line 153: Why are we passing and exploding `detector_ids`? Why not just pass the list detector_ids directly, all the detectors are being used inside `calculate_raw_data_metadata`

src/ess/nmx/mcstas/executables.py
```diff
@@ -0,0 +1,274 @@
...
+            if any(da.sizes.values()) == 0:
+                continue
+            else:
+                results = processor.add_chunk({RawEventProbability: da})
+            if logger is not None:
+                logger.info(
+                    "[{%s}/{%s}] Processed chunk for {%s}",
+                    i_da + 1,
+                    max_chunk_id,
+                    detector_name,
```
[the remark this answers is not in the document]
YooSunYoung: Yeah, ruff doesn't like it when you pre-make the string stuff.

src/ess/nmx/mcstas/executables.py
YooSunYoung: Not really a big of a reason.
I find it easier this way since we don't have to wrap one detector_id into a list or tuple in this way.

YooSunYoung:
@MridulS can you also review #115 ...?

PR #123, opened 2025-03-12
remove all zero lines, not just the first one

aaronfinke:
Changed filter function to exclude zero lines from McStas, which happen far more than only once at the beginning of the datasets, as was initially assumed. Note the performance takes a big hit due to this filtering function, but it is needed for now, until McStas is fixed. (I am not too concerned about performance.)

### DIFF
--- a/src/ess/nmx/mcstas/load.py
+++ b/src/ess/nmx/mcstas/load.py
@@ -52,8 +52,7 @@ def _exclude_zero_events(data: sc.Variable) -> sc.Variable:
     McStas can add extra event lines containing 0,0,0,0,0,0
     These lines should not be included so we skip it.
     """
-    nonzeros = data.values[(data.values != 0).all(axis=1)]
-    data = sc.array(dims=['event', 'dim_1'], values=nonzeros, unit=None)
+    data  = data[(data!=sc.scalar(0., unit=data.unit)).any(dim="dim_1")]
     return data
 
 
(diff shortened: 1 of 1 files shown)


src/ess/nmx/mcstas/load.py
```diff
@@ -49,13 +49,11 @@ def load_event_data_bank_name(
...
-    This line should not be included so we skip it.
+    McStas can add extra event lines containing 0,0,0,0,0,0
+    These lines should not be included so we skip it.
     """
-    if (data.values[0] == 0).all():
-        data = data["event", 1:]
-    else:
-        data = data
+    nonzeros = data.values[(data.values != 0).all(axis=1)]
+    data = sc.array(dims=['event', 'dim_1'], values=nonzeros, unit=None)
```
YooSunYoung, line 56: Scipp has similar function like this:
It should be faster I think...?

```suggestion
    data  = data[(data!=sc.scalar(0., unit=data.unit)).any(dim="dim_1")]
```

But some of the values can be `0` right...? So I used `any` instead of `all`.

aaronfinke: Thanks, I don't know scipp well enough right now. 

And yes, you are right, my code should have been `data.values[~(data.values == 0).all(axis=1)]`. Compared your method with mine- they're nearly the same, performance-wise. I think checking is fast, but filtering is slow. Go ahead and use your method :)

PR #125, opened 2025-03-16
add bitshuffle support, compress binned datasets using bitshuffle/LZ4

aaronfinke:
Added bitshuffle support and compression of reduced datasets with bitshuffle/lz4

### DIFF
--- a/src/ess/nmx/nexus.py
+++ b/src/ess/nmx/nexus.py
@@ -35,7 +35,7 @@ def _create_dataset_from_var(
     long_name: str | None = None,
     compression: str | None = None,
     compression_opts: int | None = None,
-    chunks: tuple[int] | int | bool | None = None,
+    chunks: tuple[int, ...] | int | bool | None = None,
     dtype: Any = None,
 ) -> h5py.Dataset:
     compression_options = {}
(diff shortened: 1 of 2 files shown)


src/ess/nmx/nexus.py
```diff
@@ -56,8 +59,8 @@ def _create_dataset_from_var(
 
 _create_compressed_dataset = partial(
     _create_dataset_from_var,
-    compression="gzip",
-    compression_opts=4,
+    compression=bitshuffle.h5.H5FILTER,
+    compression_opts=(0, bitshuffle.h5.H5_COMPRESS_LZ4),
 )
```
YooSunYoung, line 64:
```suggestion
)
"""Create dataset with compression options.

[``Bitshuffle/LZ4``](https://github.com/kiyo-masui/bitshuffle) is used for convenience. Since ``Dectris`` uses it for their Nexus file compression, it is compatible with DIALS. ``Bitshuffle/LZ4`` tends to give similar results to GZIP and other compression algorithms with better performance.

A naive implementation of bitshuffle/LZ4 compression, shown in [issue #124](https://github.com/scipp/essnmx/issues/124), led to 80% file reduction (365 MB vs 1.8 GB).

"""
```

YooSunYoung (approved):
Thank you! I just added some docstring based on your issue and then an option argument to not-compress the dataset.
We typically try not to keep the full control over this kind of side-effect/advanced settings.
But for conveniences, I kept the compression `True` as default.

YooSunYoung:
Fixes #124

aaronfinke:
in nexus.py, you write: 

> If ``True``, the detector counts are compressed using bitshuffle.
>         It is because only the detector counts are expected to be large.

This is not *necessarily* true; after all, each image is three panels of (1280,1280,50) datasets. That's ~2e8 integers. Whether those counts are all zero, the dataset will still be big. Fortunately, it is easy to compress big sets of integers like this. Bitshuffle is especially good at it.

PR #130, opened 2025-03-27
Work around bitshuffle if not available.

### LINKED ISSUE #127: bitshuffle not built/installed correctly in arm64
We probably have to test it on Windows as well but it's not working properly on Mac(arm), at least on my machine. But it seems working fine on linux.

We either not-allow bitshuffle compression unless it's linux or we find a way to use it.

YooSunYoung:
Fixes #127 

I couldn't come up with a better solution but it works for now...

### DIFF
--- a/src/ess/nmx/nexus.py
+++ b/src/ess/nmx/nexus.py
@@ -4,10 +4,9 @@ import io
 import pathlib
 import warnings
 from collections.abc import Callable, Generator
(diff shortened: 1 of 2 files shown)


jl-wynen (approved):
Should we support this in ScippNexus? It could be useful in NXCanSAS files, too. Assuming that SASView can read this.

PR #132, opened 2025-03-27
Allow -1 chunk size as a whole dataset size.

YooSunYoung:
It's cumbersome to switch between workflow for chunks and workflow for whole dataset if you don't know which one to use.

This change allows user to just load all data at once using same interface for chunk-process.

### DIFF
--- a/src/ess/nmx/mcstas/load.py
+++ b/src/ess/nmx/mcstas/load.py
@@ -121,7 +121,7 @@ def _check_maximum_chunk_size(d_slices: tuple[slice, ...]) -> None:
     _check_chunk_size(max_chunk_size)
 
 
-def _validate_chunck_size(chunk_size: int) -> None:
+def _validate_chunk_size(chunk_size: int) -> None:
     """Validate the chunk size."""
     if not isinstance(chunk_size, int):
         raise TypeError("Chunk size must be an integer.")
@@ -167,7 +167,7 @@ def raw_event_data_chunk_generator(
 
     """
     _check_chunk_size(chunk_size)
-    _validate_chunck_size(chunk_size)
+    _validate_chunk_size(chunk_size)
 
     # Find the data bank name associated with the detector
     bank_prefix = load_event_data_bank_name(
--- a/src/ess/nmx/mcstas/load.py
+++ b/src/ess/nmx/mcstas/load.py
@@ -100,6 +100,27 @@ def load_raw_event_data(
         return _wrap_raw_event_data(data)
 
 
+def _check_chunk_size(chunk_size: int) -> None:
+    if 0 < chunk_size < 10_000_000:
+        import warnings
+
+        warnings.warn(
+            "The chunk size may be too small < 10_000_000.\n"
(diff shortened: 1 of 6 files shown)


src/ess/nmx/streaming.py
```diff
@@ -85,5 +85,7 @@ def calculate_number_of_chunks(
         dset: snx.Field = root[bank_name]["events"]
         if chunk_size == 0:
             return len(list(dset.dataset.iter_chunks()))
+        elif chunk_size == -1:
+            return 1  # Read all at once
```
MridulS, line 89: should we also do a check here to make sure a negative chunk size isn't passed in like in the other file `            raise ValueError("Invalid chunk size. It should be -1(for all) or > 0.")` ? The catch all else wouldn't fail with a negative integer on a first glance.

jokasimr:
Oh didn't notice @MridulS already started a review. I'll drop the review :+1:

src/ess/nmx/streaming.py
YooSunYoung: Done..!

src/ess/nmx/mcstas/load.py
```diff
@@ -161,7 +188,8 @@ def raw_event_data_chunk_generator(
                         stacklevel=2,
```
MridulS, line 188: maybe we can call `_check_chunk_size` directly here for this warning too?

YooSunYoung: Yes. And I also refactored a bit. Thanks!

src/ess/nmx/mcstas/load.py
```diff
@@ -122,17 +151,23 @@ def raw_event_data_chunk_generator(
...
+    ValueError:
+        If the chunk size is not valid. (>= -1)
+    TypeError:
+        If the chunk size is not an integer.
+    Warning
+        If the chunk size is too small (< 10_000_000).
+
+    """
+    _check_chunk_size(chunk_size)
+    _validate_chunck_size(chunk_size)
```
MridulS, line 170:
```suggestion
    _validate_chunk_size(chunk_size)
```

src/ess/nmx/streaming.py
```diff
@@ -7,7 +7,7 @@
 
 from ess.reduce.streaming import Accumulator
 
-from .mcstas.load import load_event_data_bank_name
+from .mcstas.load import _validate_chunck_size, load_event_data_bank_name
```
MridulS, line 10:
```suggestion
from .mcstas.load import _validate_chunk_size, load_event_data_bank_name
```

src/ess/nmx/streaming.py
```diff
@@ -71,7 +71,20 @@ def calculate_number_of_chunks(
...
+
+    Raises
+    ------
+    ValueError:
+        If the chunk size is not valid. (>= -1)
+    TypeError:
+        If the chunk size is not an integer.
+
     """
+    _validate_chunck_size(chunk_size)
```
MridulS, line 87:
```suggestion
    _validate_chunk_size(chunk_size)
```

PR #134, opened 2025-03-28
Scaling dials

aaronfinke:
- added pyscale support for DIALS files directly instead of having to output as mtz
- manually inputting tof bins/max probabilities so it doesn't have to run through datasets twice;
- I have "trimmed" the mcstas output to make the nexus files smaller (much smaller, like 70%), but the relevant data is in a different form ( (n,6) array vs (n,3) ) so it tries one form and if it fails, runs the other


src/ess/nmx/dials_io.py
YooSunYoung, line 1: This file has a lot of duplication from mtz_io module but I'll just merge it for now to make the further refactoring easier.

YooSunYoung (approved):

Issue #141, opened 2025-05-21
NXLauetof (NMX specific) loader


YooSunYoung:
We can now save NXLauetof file but we don't have a loader.
The loader or the loaded object should be able to resolve the geometrical information (i.e. position vectors)
so that we can easily see the 3D instrument view from it.

PR #146, opened 2025-06-19
Reduction workflow for ess nmx files

YooSunYoung:
This is a workflow for actual nexus files collected with our filewriter.

It also has a helper to ignore empty logs but I can't guarantee that it always works.
for example, if any of the transformation is empty and one of transformation is time dependent, then the whole transformation is time dependent, therefore we can't really decide the rotation matrix.

However, it's little likely to happen.

### DIFF
--- a/tests/executable_test.py
+++ b/tests/executable_test.py
@@ -39,7 +39,11 @@ def test_executable_runs(small_nmx_nexus_path, tmp_path: pathlib.Path):
     nbins = 20  # Small number of bins for testing.
     # The output has 1280x1280 pixels per detector per time bin.
     expected_toa_bins = sc.linspace(
-        dim='dim_0', start=0, stop=int((1 / 14) * 1_000), num=nbins + 1, unit='ms'
+        dim='dim_0',
+        start=2,  # Unrealistic number for testing
+        stop=int((1 / 15) * 1_000),  # Unrealistic number for testing
+        num=nbins + 1,
+        unit='ms',
     )
     expected_toa_output = sc.midpoints(expected_toa_bins, dim='dim_0').to(unit='ns')
 


src/ess/nmx/executables.py
YooSunYoung, line 1: @aaronfinke this is the whole workflow for the real nmx nexus files.

You can use similar command line like mcstas workflow after installing this branch with pip:

```bash
pip install -e .
```

```bash
essnmx-reduce \
  --input_file docs/user-guide/977695_00052796.hdf \
  --output_file scipp_out.h5 \
  --verbose \
  --detector_ids 0 \
  --chunk_size 1_000

```

YooSunYoung: I didn't use sciline in this case or our generic workflow since it's a very simple workflow.

src/ess/nmx/nexus.py
```diff
@@ -22,6 +22,73 @@
 )
 
 
+def _fallback_compute_positions(dg: sc.DataGroup) -> sc.DataGroup:
```
YooSunYoung, line 25: This is the `hack` we need to handle the empty nxlogs.

src/ess/nmx/executables.py
```diff
@@ -0,0 +1,309 @@
...
+        fast_axis_name=fast_axis,
+        slow_axis_name='x' if fast_axis == 'y' else 'y',
+        fast_axis=fast_axis_vector,
+        slow_axis=slow_axis_vector,
+        step_x=_decide_step(da.coords['x_pixel_offset']),
+        step_y=_decide_step(da.coords['y_pixel_offset']),
+    )
+
+
+def calculate_number_of_chunks(detector_gr: snx.Group, *, chunk_size: int = 0) -> int:
```
aaronfinke, line 106: We need to check how the filewriter is writing to HDF5. If it is writing data in chunks already, we should just use that. It's more efficient

YooSunYoung: Yes we talked about implementing `iter_chunk` on scippnexus with our team.
So I'll apply it here once it's done.

Either way, we'll allow both since even the chunk might not fit in the memory for some reason...

YooSunYoung:
@aaronfinke I have tried this with the data that Doro gave me (/dmsc/scipp/nmx/cern_202506/testfile-20250625-192510.hdf)
and then found some bugs so fixed them...!

This one only has one detector so I didn't plot them in 3 D but indeed it'll be nice to be able to plot them in 3 D without Dials... So I'll prioritize #141 this one

src/ess/nmx/executables.py
aaronfinke: Getting the following error when trying to process one of Doro's files: 

```
Traceback (most recent call last):
  File "/Users/aaronfinke/miniforge3/envs/scipp/bin/essnmx-reduce", line 8, in <module>
    sys.exit(main())
             ~~~~^^
  File "/Users/aaronfinke/essnmx/src/ess/nmx/executables.py", line 387, in main
    reduction(
    ~~~~~~~~~^
        input_file=input_file,
        ^^^^^^^^^^^^^^^^^^^^^^
    ...<8 lines>...
        logger=logger,
        ^^^^^^^^^^^^^^
    )
    ^
  File "/Users/aaronfinke/essnmx/src/ess/nmx/executables.py", line 287, in reduction
    toa_bin_edges = toa_bin_edges.to(unit=da.bins.coords['event_time_offset'].unit)
  File "/Users/aaronfinke/miniforge3/envs/scipp/lib/python3.13/site-packages/scipp/core/operations.py", line 293, in to
    raise ValueError("Must provide dtype or unit or both")
ValueError: Must provide dtype or unit or both
```

YooSunYoung: Hmm... strange... I tried all three files but it was okay.

I did find some bugs so I updated this branch yesterday though. Did you pull the changes?

Also, if you need to quickly have the 2d plot, here is the code snippet:

```python
%matplotlib widget

import plopp as pp
import scippnexus as snx

with snx.File('/Users/sunyoungyoo/ESS/essnmx/testfile-20250625-192510.hdf') as f:
    dg = f['/entry/instrument/detector_panel_0/'][()]
    display(dg)
    display(dg['data'])
    da_2d = dg['data'].hist()
    da_3d = dg['data'].hist(event_time_offset=50)


display(pp.plot(da_2d, norm='log', title="2D Histogram of Detector Data"))
pp.slicer(
    da_3d,
    title="3D Histogram of Detector Data",
    norm='log',
    keep=['x_pixel_offset', 'y_pixel_offset'],
)
```

YooSunYoung:
The error was because of this: 

https://scipp.github.io/development/adr/0019-binned-data-unit-dtype-python-properties.html


I added the lower pin of scipp...

src/ess/nmx/executables.py
YooSunYoung: Mcstas executables uses StreamProcessor indeed but it's little likely that we need to process the real measurement data chunk by chunk so there was very little motivation to wrap it into a StreamProcessor.

Also the workflow itself is much more simple compared to any other workflows we have. There is more of loss than benefit for using sciline for NMX. Making it even less motivating to use StreamProcessor...

src/ess/nmx/executables.py
```diff
@@ -0,0 +1,399 @@
...
+    chunk_size: int = 1_000,
+    detector_ids: list[int | str],
+    compression: bool = False,
+    logger: logging.Logger | None = None,
+    min_toa: sc.Variable | int = 0,
+    max_toa: sc.Variable | int = int((1 / 14) * 1_000),  # Default for ESS NMX
+    toa_bin_edges: sc.Variable | int = 250,
+    fast_axis: Literal['x', 'y'] | None = None,  # 'x', 'y', or None to auto-detect
+    display: Callable | None = None,  # For Jupyter notebook display
+) -> sc.DataGroup:
```
SimonHeybrock, line 154: Definitely not a request to change anything here now, just a thought/note for the future:

In scipp/beamlime#404 I moved to using Pydantic for defining workflow parameters. As this allows for including bounds/validation, titles, and descriptions it may be a strategy we should explore more widely in the future, including when making scripts like this. It might make it simpler to document and to auto-setup and argparser, for example.

src/ess/nmx/executables.py
```diff
@@ -0,0 +1,399 @@
...
+        "--fast-axis",
+        type=str,
+        choices=['x', 'y', None],
+        default=None,
+        help="Specify the fast axis of the detector. If None, it will be determined "
+        "automatically based on the pixel offsets.",
+    )
+
+
+def main() -> None:
```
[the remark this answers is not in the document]
YooSunYoung: >  Should we try to generalize in essreduce?

Or maybe copier template...?

src/ess/nmx/executables.py
nvaytet: FYI, the error about `toa_bin_edges.to(unit=da.bins.coords['event_time_offset'].unit)` sounds like the `da.bins.coords['event_time_offset'].unit` is just returning `None`, which is why the `.to(unit=...)` is failing with that error.
My guess is that it's because you have a slightly outdated version of Scipp, where if you have binned data, you may have to do `da.bins.coords['event_time_offset'].bins.unit` instead (note the extra `.bins` before the last `.unit`).

YooSunYoung: Yeah sorry I should have mentioned this in this tread.

I already added the lower pin:  https://github.com/scipp/essnmx/pull/146#issuecomment-3012347639

src/ess/nmx/executables.py
```diff
@@ -0,0 +1,401 @@
...
+    """Reduce NMX data from a Nexus file and export to NXLauetof(ESS NMX specific) file.
+
+    This workflow is written as a flatten function without using sciline Pipeline.
+    It is because the first part of NMX reduction only requires
+    a few steps of processing and it is overkill to use a Pipeline or GenericWorkflow.
+
+    We also do not apply frame unwrapping or pulse skipping here,
+    as it is not expected from NMX experiments.
+
+    Frame unwrapping may be applied later on the result of this function if needed
```
nvaytet: If you want to defer the time-of-flight calculation to a later PR, we should at least amend this comment to say that we still need to implement tof computation.
If not, this will just get forgotten.

PR #155, opened 2025-11-10
Retrieve crystal rotation from nexus file.

YooSunYoung:
We need to export crystal_rotation (sample orientation) to the output file for next step of reduction.
We haven't settled about how exactly to store this information in the real nexus file.
This is a improvised temporary solution for (mcstas to nexus) translated files.

### DIFF
--- a/src/ess/nmx/executables.py
+++ b/src/ess/nmx/executables.py
@@ -30,6 +30,24 @@ def _retrieve_sample_position(file: snx.File) -> sc.Variable:
     return _compute_positions(da, auto_fix_transformations=True)['position']
 
 
+def _retrieve_crystal_rotation(file: snx.File) -> sc.Variable:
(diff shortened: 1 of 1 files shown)


src/ess/nmx/executables.py
```diff
@@ -30,6 +30,24 @@ def _retrieve_sample_position(file: snx.File) -> sc.Variable:
...
+    if 'crystal_rotation' not in file['entry/sample']:
+        import warnings
+
+        warnings.warn(
+            "No crystal rotation found in the Nexus file under "
+            "'entry/sample/crystal_rotation'. Returning zero rotation.",
+            RuntimeWarning,
+            stacklevel=2,
+        )
+        return sc.vector([0, 0, 0], unit='deg')
```
jokasimr: Do we need this for backwards compatibility?

jokasimr (approved):

PR #160, opened 2025-11-17
Organize reduction configuration using pydantic model

YooSunYoung:
These updates have following changes:

- Update command arguments name to have more clear meaning
- Update help messages to explain each arguments better
- Use Pydantic model to build configuration before using it in the reduction
- Allow pydantic model instance to dump command line arguments so that users can easily copy them.
- Add `swrm` mode argument as mentioned in scipp/ess#196 (not entirely implemented yet)
- Add unit tests for pydantic model/command argument parsing

**Not in the scope works**
- Actually using configuration values in the reduction flow

### DIFF
--- a/src/ess/nmx/_executable_helper.py
+++ b/src/ess/nmx/_executable_helper.py
@@ -189,7 +189,16 @@ class ReductionConfig(BaseModel):
     def _children(self) -> list[BaseModel]:
         return [self.inputs, self.workflow, self.output]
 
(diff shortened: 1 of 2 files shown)


SimonHeybrock (commented):
Have you considered using `pydantic.Field` with `title` and `description` options, like we do in ESSlivedata? Then those things are in a central place and it will be compatible with others, instead of having, e.g., the description hidden (and potentially duplicated) when building the argparser?

PR #162, opened 2025-11-19
Use pydantic field to automate argument building and parsing.

YooSunYoung:
Continued from #160

### DIFF
--- a/src/ess/nmx/_executable_helper.py
+++ b/src/ess/nmx/_executable_helper.py
@@ -18,10 +18,6 @@ from .types import Compression
 
 
 def _validate_annotation(annotation) -> TypeGuard[type]:
-    # Supported annotation for command arguments:
-    # - Atomic types: int, float, str, bool, enum.StrEnum, Literal
-    # - Optional[AtomicType]
-    # - List[AtomicType], Tuple[AtomicType, ...], Set[AtomicType]
     def _validate_atomic_type(annotation) -> bool:
         return (
             (annotation in (int, float, str, bool))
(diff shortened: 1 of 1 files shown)


src/ess/nmx/_executable_helper.py
```diff
@@ -5,168 +5,217 @@
...
+    field_name: str, field_info: FieldInfo, args: argparse.Namespace
+):
+    if isinstance(field_info.annotation, type) and issubclass(
+        field_info.annotation, Enum
+    ):
+        return field_info.annotation[getattr(args, field_name)]
+    return getattr(args, field_name)
+
+
+class CommandArgument(BaseModel):
```
SimonHeybrock: Not sure why this needs to be used as a base model. Can't the same be done with a free function such that models do not need to inherit?

src/ess/nmx/_executable_helper.py
```diff
@@ -5,168 +5,217 @@
...
-        return parser
 
-    @classmethod
-    def from_args(cls, args: argparse.Namespace) -> "OutputConfig":
-        return cls(
-            verbose=args.verbose,
-            output_file=args.output_file,
-            compression=Compression[args.compression],
-        )
+class TOAUnit(Enum):
```
SimonHeybrock: Consider `StrEnum`, since we now have Python 3.11

src/ess/nmx/_executable_helper.py
```diff
@@ -5,168 +5,217 @@
...
+        description="Minimum time of arrival (TOA) in [toa_unit].",
+        default=0,
+    )
+    max_toa: int = Field(
+        title="Maximum Time of Arrival",
+        description="Maximum time of arrival (TOA) in [toa_unit].",
+        default=int((1 / 14) * 1_000),
+    )
+    toa_unit: TOAUnit = Field(
+        title="Maximum Time of Arrival",
```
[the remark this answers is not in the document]
YooSunYoung: Fixed it! But which titles you mean? It was only toa_unit...

src/ess/nmx/_executable_helper.py
```diff
@@ -1,175 +1,244 @@
...
+from pydantic_core import PydanticUndefined
 
 from .types import Compression
 
 
+def _validate_annotation(annotation) -> TypeGuard[type]:
+    # Supported annotation for command arguments:
+    # - Atomic types: int, float, str, bool, enum.StrEnum, Literal
+    # - Optional[AtomicType]
+    # - List[AtomicType], Tuple[AtomicType, ...], Set[AtomicType]
```
SimonHeybrock: Maybe this belong into the docstring of `add_args_from_pydantic_model`?

YooSunYoung: I added docstring to the helper functions including these lines.

PR #167, opened 2025-12-04
Update configurations and reduction interface function signature.

YooSunYoung:
- Add tof simulation seed to the configuration.
- Use configuration object instead of passing individual argument.

### DIFF
--- a/src/ess/nmx/executables.py
+++ b/src/ess/nmx/executables.py
@@ -234,14 +234,18 @@ def reduction(
 
     """
     display = _retrieve_display(logger, display)
+    input_file_path = _retrieve_input_file(
+        collect_matching_input_files(*config.inputs.input_file)
+    ).resolve()
+    display(f"Input file: {input_file_path}")
+
+    output_file_path = pathlib.Path(config.output.output_file).resolve()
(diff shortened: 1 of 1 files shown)


src/ess/nmx/executables.py
```diff
@@ -190,57 +192,26 @@ def _retrieve_input_file(input_file: list[pathlib.Path] | pathlib.Path) -> pathl
...
-    compression: Compression = Compression.BITSHUFFLE_LZ4,
+    config: ReductionConfig,
     logger: logging.Logger | None = None,
-    min_toa: sc.Variable | int = 0,
-    max_toa: sc.Variable | int = int((1 / 14) * 1_000),  # Default for ESS NMX
-    toa_bin_edges: sc.Variable | int = 250,
-    fast_axis: Literal['x', 'y'] | None = None,  # 'x', 'y', or None to auto-detect
-    display: Callable | None = None,  # For Jupyter notebook display
+    display: Callable | None = None,
 ) -> sc.DataGroup:
```
YooSunYoung: This is the only important part. 
The rest of the function will be updated in the later PR.

PR #172, opened 2025-12-05
Add documentation of reduction interface and allow skipping file saving.

YooSunYoung:


src/ess/nmx/configurations.py
```diff
@@ -152,3 +163,45 @@ class ReductionConfig(BaseModel):
     @property
     def _children(self) -> list[BaseModel]:
         return [self.inputs, self.workflow, self.output]
+
+
+def to_command_arguments(
```
YooSunYoung: I moved this interface to the `configurations` module so that users can use it.

PR #201, opened 2026-02-25
Nxlauetof exporter and loader update for easier geometry reconstruction.

YooSunYoung:
It was a bit hard to validate the result without custom loader for the NMX output file (nxlauetof).

Most changes is related to adjusting the in-memory result data structure, so it should not affect the output file.
Output files are meant to be read by DIALS and it won't use this loader so regression tests won't be necessary.
The loader is only for the IDS member and developers.

The loader adjust and reconstruct some coordinates from the file.
And there is a test that checks if the reconstructed positions are good enough (i.e. difference less than 10th of pixel size.)


YooSunYoung:
@aaronfinke  I found the wrong calculation in the code and fixed it...! Now it gives the correct position.
It still has a very small difference of `4.209804057220788e-11 mm` from the original positions in distance.
It is using `origin` position instead of the `first_pixel_position`.

src/ess/nmx/types.py
```diff
@@ -118,10 +195,71 @@ class NMXDetectorMetadata:
...
+
+    control: NMXMonitorMetadata
+    definitions: Literal['NXlauetof'] = 'NXlauetof'
+    instrument: NMXInstrument
+    sample: NMXSampleMetadata
+    lookup_table: TofLookupTable | None = None
+    reducer: NMXProgram = field(default_factory=NMXProgram)
+    "Information of the reduction software."
+
+    def to_datagroup(self) -> sc.DataGroup:
```
jl-wynen: If this is just for display, can you turn this into `_repr_mimebundle_`?

docs/user-guide/workflow.ipynb
```diff
@@ -52,8 +52,30 @@
...
+    "filename = data.get_path('PG3_4844_event.nxs')\n"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "!cp '/home/sunyoungyoo/.cache/scippnexus/1/PG3_4844_event.nxs' ./"
```
jl-wynen: Please remove this and use portable code instead.

docs/user-guide/workflow.ipynb
```diff
@@ -149,6 +171,57 @@
...
    ]
   },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Loading Reduced File\n",
+    "\n",
+    "There is a custom loader for NXlauetof file for NMX.<br>\n",
+    "It reconstruct the position coordinates from the file and add it back to the data array.<br>\n",
```
jl-wynen:
```suggestion
    "It reconstructs the position coordinates from the file and adds them back to the data array.<br>\n",
```

jl-wynen (approved):

PR #203, opened 2026-03-02
deps: lower bound numpy

jokasimr:
Fixes failing nightly lower bound test


pyproject.toml
```diff
@@ -44,6 +44,7 @@ dependencies = [
   "defusedxml>=0.7.1",
   "msgpack>=1.0.8",
   "tof>=25.12.1",
+  "numpy>=1.20.0",
```
YooSunYoung:
```suggestion
  "numpy>=2.0.0",
```

It still fails, https://github.com/scipp/essnmx/actions/runs/22568135904/job/65368889178

I think we can set the lower bound of numpy 2.
essnmx is more of an application than a library.

YooSunYoung (approved):

PR #206, opened 2026-04-01
'definitions' to 'definition'

aaronfinke:
I'm pretty sure this is a typo in `nexus.py`... `/entry/definitions` should be `/entry/definition`


nvaytet:
@YooSunYoung found this one in my backlog. Do you agree with the changes?

